PR #7, opened 2017-06-25
Add ax_cxx_compile_stdcxx(.m4) file for build with Ubuntu 16.04

alagoutte:
to avoid
./configure: line 17124: syntax error near unexpected token AX_CXX_COMPILE_STDCXX(14, noext, optional)'

When ./configure


tatsuhiro-t:
Thank you.  I forgot to add it.

PR #8, opened 2017-06-25
Add note about need libevent for build examples

alagoutte:
(May be also add check if libevent is available because the build fail :
In file included from client.cc:38:0:
client.h:38:16: fatal error: ev.h: No such file or directory

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,6 +38,10 @@ libngtcp2 uses cunit for its unit test frame work:
 
 * cunit >= 2.1
 
+To build sources under the examples directory, libev is required:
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you.
I think it is libev, not libevent-openssl.

alagoutte:
i based from nghttp2 ;-) (and get libev and libevent-openssl...)

alagoutte:
Updated the patch

tatsuhiro-t:
Merged now.  Thank you!

PR #10, opened 2017-06-27
Add Travis CI

alagoutte:
Need to use clang-3.8 and gcc5 for c++std14


alagoutte:
Not enable -werror because get some build error :)

tatsuhiro-t:
Thank you!  Merged now.
I fixed gcc compile error with --enable-werror. It should be fine to enable it in travis build.

PR #11, opened 2017-06-29
debug: Display PKN for Packet Number like Wireshark

alagoutte:
(and also from GQUIC...


alagoutte:
on first (Google)Draft, it is PKN for Packet Number... 

What do you think ?

tatsuhiro-t:
I can live with PKN.

PR #12, opened 2017-06-29
Remove last nghttp2 trace...

alagoutte:
No sure for header

but there is a typo on ngtcp2.h comment (about mem)

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -125,7 +125,7 @@ typedef void *(*ngtcp2_realloc)(void *ptr, size_t size, void *mem_user_data);
  *     }
  *
  *     void conn_new() {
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR!
For copyright notice, I would rather keep them as is.  This is because I copied code from nghttp2 project, and they are licensed under MIT with those contributors.
If this is not normal for opensource project, I have all ears, and I'll accept your patch as is.

alagoutte:
i will update the patch for only fix ngtcp2.h typo about mem

May be add a note somewhere for say, it is a project based on some code from nghttp2 ?

tatsuhiro-t:
Thank you.  I thought that simply showing copyright is OK .

alagoutte:
Update to fix only typo on ngtcp2.h file

tatsuhiro-t:
Thank you!

PR #17, opened 2017-07-05
docs: Add note about need golang

alagoutte:
BoringSSL need golang for build


tatsuhiro-t:
Thank you.  I completely forget this fact.

PR #27, opened 2017-07-18
Transport params

tatsuhiro-t:
Send and receive TransportParameters.


tatsuhiro-t:
CI error is due to missing API in boringssl.  That's ok because we will drop boringssl support soon.

PR #29, opened 2017-07-25
Add make gen-certificate for generate DEMO certificate

alagoutte:


tatsuhiro-t:
Thank you.  Merged now.

Lekensteyn:
What is this used for? Is the password really necessary? Any reason why it has to be a Makefile target, and not a script like `ci/gen-certificate.sh`?

If I ever need a RSA key and self-signed certificate for testing, I normally create one server.pem file with both files:
```
openssl req -newkey rsa:2048 -x509 -nodes -keyout server.pem -new -out server.pem -subj /CN=localhost
```

Add `-days 30` as desired, or `-sha256` for such signatures and change `-nodes` if you do want password-protected keyfiles.

alagoutte:
Hi @Lekensteyn 

The passwordis not really necessary
ci/gen-certificate(.sh) is good idea...

PR #30, opened 2017-07-25
travis: use also ./config for build

alagoutte:


tatsuhiro-t:
Thank you!  Merged now.

PR #31, opened 2017-07-26
Invalid memory access in Client Server examples

nickfajones:
The ngtcp2_conn_callbacks and ngtcp2_settings structures that
are passed into ngtcp2_conn_{client,server}_new only exist in
the scope of the stack that the above function is called.

Once the function returns these objects no longer exist yet
they continue to be referenced throughout the lifetime of
the conn.

Fix this by making the callback and settings structures into
members of the Client and (Server) Handler classes.

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -188,6 +188,8 @@ Client::Client(struct ev_loop *loop, SSL_CTX *ssl_ctx)
       loop_(loop),
       ssl_ctx_(ssl_ctx),
       ssl_(nullptr),
+      callbacks_{},
+      settings_{},
       fd_(-1),
       stdinfd_(-1),
       stream_id_(0),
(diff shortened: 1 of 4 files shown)


tatsuhiro-t:
They are allocated on the stack, but in side ngtcp2_conn_server_new, they are copied by value after dereference.  https://github.com/ngtcp2/ngtcp2/blob/9a38e10166e591278a7c49ff9dd67967e2adeb1b/lib/ngtcp2_conn.c#L188
So I think the current implementation is safe.  Have you actually experienced any memory violation?

nickfajones:
Understood, I misread the code.  As you pointed out, the entire structure is copied during the assignment, so lifetime of the parameter object is no longer relevant.  Sorry for the trouble!

Closing request.

PR #35, opened 2017-08-04
Improve ease of debugging of client/server examples

nickfajones:
- Define new CFLAGS in debug mode:
  - enable full embedding of debugging symbols
  - lower optimisation to avoid 'jumping' in gdb
  - enable stricter compiler warnings
- mitigate compiler warnings in client example
- Build client and server executables as binaries to
  allow them to be invoked directly by gdb
- Activiate the --enable-werror flag
- Handle the case of slightly overlapping werror
  cflags and debug cflags


configure.ac
```diff
@@ -63,7 +63,13 @@ AC_ARG_ENABLE([werror],
 AC_ARG_ENABLE([debug],
     [AS_HELP_STRING([--enable-debug],
                     [Turn on debug output])],
-    [debug=$enableval], [debug=no])
+    [debug=yes], [debug=no])
```
tatsuhiro-t: I think debug=$enableval is fine.  If we do debug=yes here, it breaks --disable-debug.

configure.ac
```diff
@@ -63,7 +63,13 @@ AC_ARG_ENABLE([werror],
...
+
+if test "x${debug}" = "xyes"; then
+  CFLAGS="-Wall -Werror -Wfatal-errors -O0 -g3"
+fi
+
+AM_CONDITIONAL([DEBUG], [test "x${debug}" = "xyes"])
```
tatsuhiro-t: It looks like we don't use DEBUG in Makefile.am, so perhaps we don't need this.

configure.ac
```diff
@@ -63,7 +63,13 @@ AC_ARG_ENABLE([werror],
 AC_ARG_ENABLE([debug],
     [AS_HELP_STRING([--enable-debug],
                     [Turn on debug output])],
-    [debug=$enableval], [debug=no])
+    [debug=yes], [debug=no])
+
+if test "x${debug}" = "xyes"; then
+  CFLAGS="-Wall -Werror -Wfatal-errors -O0 -g3"
```
tatsuhiro-t: As far as I know, changing CFLAGS user variable here will break user's ability to change CFLAGS, and it is discouraged.
I think it is better to define new variable, say, DEBUGCFLAGS, and append it to AM_CFLAGS in various Makefile.am.  The existing example is WARNCFLAGS.

tatsuhiro-t:
Thank you for PR.  Could you check out my comments?

configure.ac
nickfajones: Will do...

configure.ac
nickfajones: Will do...

configure.ac
nickfajones: Fighting with autoconf: I need to find a way to suppress the autoconf default CFLAGS, but only if they are default, otherwise let user defined CFLAGS through.  Unfortunately autoconf doesn't differentiate between default and user supplied CFLAGS, so I need to find an elegant way to do this myself...

nickfajones:
Not sure if you were notified, but I updated my tree and it seems this PR was automatically updated as well

configure.ac
```diff
@@ -54,17 +54,37 @@ PACKAGE_VERSION_NUM=`printf "0x%02x%02x%02x" "$major" "$minor" "$patch"`
...
+  WERRORCFLAGS="$_WERRORCFLAGS"
+  if test "x${debug}" = "xno"; then
+# only perform actual substitution if debug is not enabled
+# the following dummy variable declaration is needed to
+# work around an autoconf behaviour: AC_SUBST will not work
+# without a (any) variable declaration in front of it
+    _DUMMY=
+    AC_SUBST([WERRORCFLAGS])
+  fi
+fi
```
tatsuhiro-t: I think we have already handled werror later in the configure.ac (WARNCFLAGS).  So we don't have to add extra stuff here.

tatsuhiro-t:
Thank you.  Merged now.

PR #36, opened 2017-08-08
Fix error caused by C warning flags propogating to C++

nickfajones:
My sincere apologies for breaking builds and Continuous Integration
with my last PR.  I failed to properly check the --enable-werror flag and
found that my change was passing all of these warning flags to C++
build targets, and C++ would choke on some of these options.

To fix this, I have:
- Stopped propogating warning flags to C++ build targets: I will look at
  defining a specific set of flags for C++ in future
- Trimmed DEBUGFLAGS down to just optimisation and debug level
  flags

Sorry, I will be more careful in future.


tatsuhiro-t:
No worries.  Thank you for contributing to the project!

PR #37, opened 2017-08-10
Add a counter to rtb_entry

nickfajones:
Count the number of times a packet has been retransmitted.
At this stage, it's only purpose is to slowly backoff
retransmit timeouts so as not to cause flooding.
In future it could be used for statistics reporting and
telemetry, especially if a similar counter is kept at
the connection level.

Note: a good way to observe retransmit behaviour before
and after this change is to put a gdb breakpoint at
'ngtcp2_conn_recv' in the example client then run the
client and observe the already running server.


tatsuhiro-t:
Thank you!  Merged now.

PR #38, opened 2017-08-15
Make memory leak checking easier by adding cleanups

nickfajones:
Some static objects used by the client and server
examples need cleanups so that they don't add noise
to valgrind output, and for the sake of correctness:
- Allow static BIOMETHOD struct to be cleaned up at
  exit
- Add defered cleanup for the default ev_loop
- Add a signal handler so that example app exits
  gracefully instead of dying

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -133,9 +133,8 @@ int bio_destroy(BIO *b) {
 }
 } // namespace
 
-namespace {
-BIO_METHOD *create_bio_method() {
-  static auto meth = BIO_meth_new(BIO_TYPE_FD, "bio");
+BIOMethod::BIOMethod()
+    : meth(BIO_meth_new(BIO_TYPE_FD, "bio")) {
   BIO_meth_set_write(meth, bio_write);
   BIO_meth_set_read(meth, bio_read);
   BIO_meth_set_puts(meth, bio_puts);
@@ -143,8 +142,15 @@ BIO_METHOD *create_bio_method() {
   BIO_meth_set_ctrl(meth, bio_ctrl);
   BIO_meth_set_create(meth, bio_create);
   BIO_meth_set_destroy(meth, bio_destroy);
-  return meth;
 }
+
+BIOMethod::~BIOMethod() {
+  BIO_meth_free(meth);
+  meth = nullptr;
+}
+
+namespace {
+BIOMethod biomethod;
 } // namespace
 
 namespace {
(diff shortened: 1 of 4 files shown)


tatsuhiro-t:
Thank you for the PR.
In my opinion, it is not so bad not to delete statically allocated object at exit.  I don't use valgrind these days, but ASAN does not say anything at the current form of the implementation.  Also it avoids some problems: https://isocpp.org/wiki/faq/ctors#static-init-order

For signal handlers, I think it is ok to just exit on SIGINT.  Are you caring about "leak" even if we hit ctrl-C?

nickfajones:
I am not familiar with ASAN and how it reports leaked memory, but valgrind does report static object leaks, so when the valgrind summary appears, it is frustrating to see page after page of leak reports for these static objects, I just wanted to clean these up to allow us to focus on legitimate leaks when using valgrind.  Also as I mentioned in the comment, I feel it is the correct thing to do.

I agree with your point about init ordering, in fact it gets worse than that: because I have found objects in the **same** compilation unit may have unpredictable ordering (possibly due to C++ variable name mangling?), so for ordering enforcement I tend to rely on: `__attribute__(init_priority(***))`

Construct on first use certainly has its advantages but I avoid it out of habit because of how this method behaves in multi threaded systems, unless you bundle in a static mutex as well, and of course static construct on first use doesn't have any regard for cleanup.

Let me experiment with this to find a solution that is the best of both worlds, I will try to create a templated wrapper that can perform static construct on first use of an object and will also support destruction (much like your `defer` template), as well as providing optional thread safety built in.

As for SIGINT (and SIGTERM for daemons) handling:  I think it is essential that any application properly unwind itself when it is told to shut down, even if it is just an example app.

This is not just for the sake of cleaning up allocated memory, there may also be some shutdown procedures that need to be followed, in fact there is one shutdown procedure that is currently missing from ngtcp2: at application stop time, the library needs to iterate over all open connections and move these into shutdown state where each connection sends: GOAWAY then CONNECTION_CLOSE to its opposite endpoint.  Both the example apps and the core library should have interfaces to support this.  So I hope you will consider the signal handler! :-)  Do you want me to split signal handling into a different PR?

tatsuhiro-t:
Thanks.  I tend to avoid those C/C++ semantics by just "leaking" them, but I'm fine to wait for you to experiment it.

You are right about connection close procedure.  I was in the mind set of TCP, and forget the fact that OS does the connection closure, while in QUIC, we have to do it ourselves.
It looks like that GOAWAY is removed from transport doc.  Just sending CONNECTION_CLOSE is enough for every connection.  I prefer the separate PR for this connection closure if it is possible.

nickfajones:
OK, so I will close this PR and in its place:
- One PR just for SIGINT handling
- One PR for a 'better' approach to static initialisation (later)
- One Issue describing a procedure for closing all connections

PR #39, opened 2017-08-16
Add SIGINT handler to example apps

nickfajones:
This enables a graceful app exit and in future will allow:
- Closure procedures like CONNECTION_CLOSE on all connections
  to be performed
- Full cleanup of allocated memory and resources


tatsuhiro-t:
Thank you.  Merged now.

PR #41, opened 2017-08-22
Close all connections on exit (Phase 1)

nickfajones:
When an example application exits, iterate over all client
and server connections and send a CONNECTION_CLOSE message
on each before cleaning up all resources.

This is only a simple initial implementation (Phase 1)
that will successfully close a small number of connections.
Phase 2 will deal with larger numbers of connections and
the load that closing these will place on a single UDP
socket.

----

Hello Tatsuhiro, this is only phase 1 of this feature.  It works, but
only for our simple example client and server applications.
A more complete solution (phase 2) will be submitted as a
future PR.

The reason why a phase 2 is needed is this: for an application
with more connections, for example a busy server, I predict
that sending a large number of CONNECTION_CLOSE
packets will cause the network layer tx buffer to fill up,
particularly if the application was busy just prior to shutdown.

When this happens, the `sendmsg` system call will probably
return `-1` with `errno` set to `EAGAIN`.  At this time we will
need to enter a final event loop, observing the write side ev_io
watcher (as well as a deadline timer), continue the cycle of:
`((UDP write)* + io_wait)*` until all connections are closed,
then exit, unless the deadline timeout is hit, then we just give
up.

So in phase 2, the connection close procedure will be split
into two phases: `disconnect`, `disconnect_continue` and
`close`.  Currently in phase 1, I have split: `disconnect` and
`close`, as an indicator of what I intend to do.

Sorry to be so quiet lately, I was on holiday.

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -184,8 +184,6 @@ void timeoutcb(struct ev_loop *loop, ev_timer *w, int revents) {
   debug::print_timestamp();
   std::cerr << "Timeout" << std::endl;
 
-  c->handle_error(0);
-
   c->disconnect();
 }
 } // namespace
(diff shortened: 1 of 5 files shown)


tatsuhiro-t:
Looks good.  Thank you!
Thank you for pointing out the closing phase.  Yes, we have to handle the situation.
It may be dealt with pacing packet when we are implementing loss recovery and congestion control.

PR #42, opened 2017-08-25
Add packet send error checking and retry

nickfajones:
This patch represents phase 2 of the "Close all
connections on exit" feature, as well as including
a number of other benefits for both client and server:

- Packet writing is now moved to a function: send_packet
  this function will sense a number of different errors
  on the sendto system call, and categorise these as
  either fatal or non fatal (able to be retried).
  Non fatal errors will most likely be caused by the
  kernel tx queue being full, which means the packet
  should be writable at a later time.
  This code is loosely based on similar code in the
  Berkeley Internet Name Daemon (BIND), which I
  assume would know how to send UDP packets.
- Functions that write packets will observe the
  send_packet return value and if the network error is
  non fatal, the packet that was to be sent will be
  queued in the backend and write action on the
  socket file descriptor will be polled through the
  event reactor.
- When writes are once again possible on the socket,
  queued packets will be retried first, to preserve
  ordering at the send side, and only once these
  packets are all sent will the core library be queried
  for more new transport level and application level
  packets, which themselves may need to be queued.

----

Hello Tatsuhiro, this change is both a Phase 2 and a
Phase 1.  Firstly, it is a Phase 2 as it addresses the
issue of retrying packets when writes to the socket
fail in non-fatal ways, as I discussed in the previous
PR, however, it is also a Phase 1 because there is
much more to do at this level:  This PR uses a basic
queue to store packets, however I believe there is
much more information that we need to store along
with the queued packet data, and there are also some
changes that need to be made to the packet writing
process:

- I believe the packet header, and the pointer to the 
  ngtcp2_conn need to be stored in a struct along
  with the packet bytes.  In addition, there probably
  needs to be some kind of retry counter or timeout
  to fail a packet if it takes too long to send.
- Currently, when a packet is written, it is constructed
  using the ngtcp2_upe_final or ngtcp2_ppe_final
  function, then immediately following that, the rtb
  entry is created and stored.  Only after that does
  the function return so that the packet can be
  written to the UDP socket.  I believe that the order
  of these operations should be changed:
- The UDP write should be called after packet
  construction but before rtb registration.  This
  could be done as another ngtcp2_conn_callback:
  (write_pkt), and this is where the UDP packet will
  be written (like the send_packet function I added
  n this PR)
- If the return value of this write_pkt callback
  function is OK, then the core library will continue
  to create the rtb.  If there is a non-fatal error, the
  library will queue the packet and other information,
  but not create the rtb, as we have not written the
  UDP packet yet, therefore shouldn't start the
  timer for its ACK.

What this will achieve is: the movement of more
responsibility from the example applications and
into the core library (the transport layer), as I'm
sure you agree that the client and server examples
are doing a too much.  However, I didn't want to go
that far in this PR in order to keep the size of the
change under control.

Please let me know what you think.  Thanks!

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -60,9 +60,9 @@ constexpr size_t MAX_BYTES_IN_FLIGHT = 1460 * 10;
 } // namespace
 
 Buffer::Buffer(const uint8_t *data, size_t datalen)
-    : buf{data, data + datalen}, pos(std::begin(buf)) {}
-
-Buffer::Buffer() : pos(std::begin(buf)) {}
+    : buf{data, data + datalen}, head(0), tail(datalen) {}
+Buffer::Buffer(size_t datalen) : buf(datalen), head(0), tail(0) {}
+Buffer::Buffer() : head(0), tail(0) {}
 
 namespace {
 int bio_write(BIO *b, const char *buf, int len) {
@@ -149,10 +149,14 @@ BIO_METHOD *create_bio_method() {
 
 namespace {
 void writecb(struct ev_loop *loop, ev_io *w, int revents) {
+  ev_io_stop(loop, w);
+
   auto c = static_cast<Client *>(w->data);
 
-  if (c->on_write() != 0) {
+  auto rv = c->on_write();
+  if (rv == NETWORK_ERR_SEND_FATAL) {
     c->disconnect();
+    return;
   }
 }
 } // namespace
@@ -162,6 +166,11 @@ void readcb(struct ev_loop *loop, ev_io *w, int revents) {
   auto c = static_cast<Client *>(w->data);
 
   if (c->on_read() != 0) {
+    c->disconnect();
+    return;
+  }
+  auto rv = c->on_write();
+  if (rv == NETWORK_ERR_SEND_FATAL) {
     c->disconnect();
   }
 }
@@ -219,6 +228,7 @@ Client::Client(struct ev_loop *loop, SSL_CTX *ssl_ctx)
       conn_(nullptr),
       crypto_ctx_{},
       streambuf_idx_(0),
+      sendbuf_{NGTCP2_MAX_PKTLEN_IPV4},
       tx_stream_offset_(0),
       should_send_fin_(false) {
   ev_io_init(&wev_, writecb, 0, EV_WRITE);
@@ -566,40 +576,48 @@ int Client::on_read() {
 
   ev_timer_again(loop_, &timer_);
 
-  return on_write();
+  return 0;
 }
 
 int Client::on_write() {
-  std::array<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> buf;
-  assert(buf.size() >= max_pktlen_);
+  if (sendbuf_.size() > 0) {
+    auto rv = send_packet();
+    if (rv != NETWORK_ERR_OK) {
+      return rv;
+    }
+  }
 
   for (auto it = std::begin(streambuf_) + streambuf_idx_;
        it != std::end(streambuf_); ++it) {
     auto &v = *it;
-    if (on_write_stream(stream_id_, 0, v) != 0) {
-      return -1;
+    auto rv = on_write_stream(stream_id_, 0, v);
+    if (rv != 0) {
+      return rv;
     }
-    if (v.left() > 0) {
+    if (v.size() > 0) {
       break;
     }
     ++streambuf_idx_;
   }
 
   if (streambuf_idx_ == streambuf_.size() && should_send_fin_) {
-    should_send_fin_ = false;
     auto v = Buffer{};
-    if (on_write_stream(stream_id_, 1, v) != 0) {
-      return -1;
+    auto rv = on_write_stream(stream_id_, 1, v);
+    if (rv != 0) {
+      return rv;
     }
+    should_send_fin_ = false;
   }
+
+  assert(sendbuf_.left() >= max_pktlen_);
 
   for (;;) {
     ssize_t n;
     if (ngtcp2_conn_bytes_in_flight(conn_) < MAX_BYTES_IN_FLIGHT) {
-      n = ngtcp2_conn_write_pkt(conn_, buf.data(), max_pktlen_,
+      n = ngtcp2_conn_write_pkt(conn_, sendbuf_.wpos(), max_pktlen_,
                                 util::timestamp());
     } else {
-      n = ngtcp2_conn_write_ack_pkt(conn_, buf.data(), max_pktlen_,
+      n = ngtcp2_conn_write_ack_pkt(conn_, sendbuf_.wpos(), max_pktlen_,
                                     util::timestamp());
     }
     if (n < 0) {
@@ -611,15 +629,14 @@ int Client::on_write() {
       break;
     }
 
-    if (debug::packet_lost(config.tx_loss_prob)) {
-      std::cerr << "** Simulated outgoing packet loss **" << std::endl;
-      continue;
-    }
+    sendbuf_.push(n);
 
-    auto nwrite = write(fd_, buf.data(), n);
-    if (nwrite == -1) {
-      std::cerr << "write: " << strerror(errno) << std::endl;
-      return -1;
+    auto rv = send_packet();
+    if (rv == NETWORK_ERR_SEND_NON_FATAL) {
+      break;
+    }
+    if (rv != NETWORK_ERR_OK) {
+      return rv;
     }
   }
 
@@ -628,8 +645,6 @@ int Client::on_write() {
 }
 
 int Client::on_write_stream(uint32_t stream_id, uint8_t fin, Buffer &data) {
-  std::array<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> buf;
-  assert(buf.size() >= max_pktlen_);
   size_t ndatalen;
 
   for (;;) {
@@ -637,9 +652,9 @@ int Client::on_write_stream(uint32_t stream_id, uint8_t fin, Buffer &data) {
       break;
     }
 
-    auto n = ngtcp2_conn_write_stream(conn_, buf.data(), max_pktlen_, &ndatalen,
-                                      stream_id, fin, data.rpos(), data.left(),
-                                      util::timestamp());
+    auto n = ngtcp2_conn_write_stream(conn_, sendbuf_.wpos(), max_pktlen_,
+                                      &ndatalen, stream_id, fin, data.rpos(),
+                                      data.size(), util::timestamp());
     if (n < 0) {
       switch (n) {
       case NGTCP2_ERR_STREAM_DATA_BLOCKED:
@@ -654,23 +669,16 @@ int Client::on_write_stream(uint32_t stream_id, uint8_t fin, Buffer &data) {
       return -1;
     }
 
-    data.pos += ndatalen;
+    data.seek(ndatalen);
 
-    if (debug::packet_lost(config.tx_loss_prob)) {
-      std::cerr << "** Simulated outgoing packet loss **" << std::endl;
-      if (data.left() == 0) {
-        break;
-      }
-      continue;
+    sendbuf_.push(n);
+
+    auto rv = send_packet();
+    if (rv != NETWORK_ERR_OK) {
+      return rv;
     }
 
-    auto nwrite = write(fd_, buf.data(), n);
-    if (nwrite == -1) {
-      std::cerr << "write: " << strerror(errno) << std::endl;
-      return -1;
-    }
-
-    if (data.left() == 0) {
+    if (data.size() == 0) {
       break;
     }
   }
@@ -706,8 +714,8 @@ size_t Client::read_client_handshake(const uint8_t **pdest) {
     return 0;
   }
   const auto &v = chandshake_[chandshake_idx_++];
-  *pdest = v.buf.data();
-  return v.buf.size();
+  *pdest = v.rpos();
+  return v.size();
 }
 
 size_t Client::read_server_handshake(uint8_t *buf, size_t buflen) {
@@ -808,6 +816,38 @@ ssize_t Client::decrypt_data(uint8_t *dest, size_t destlen,
 
 ngtcp2_conn *Client::conn() const { return conn_; }
 
+int Client::send_packet() {
+  if (debug::packet_lost(config.tx_loss_prob)) {
+    std::cerr << "** Simulated outgoing packet loss **" << std::endl;
+    return NETWORK_ERR_OK;
+  }
+
+  int eintr_retries = 5;
+  ssize_t nwrite = 0;
+
+  do {
+    nwrite = sendto(fd_, sendbuf_.rpos(), sendbuf_.size(), 0,
+                    &remote_addr_.su.sa, remote_addr_.len);
+  } while ((nwrite == -1) && (errno == EINTR) && (eintr_retries-- > 0));
+
+  if (nwrite == -1) {
+    switch (errno) {
+    case EAGAIN:
+    case EINTR:
+    case 0:
+      return NETWORK_ERR_SEND_NON_FATAL;
+    default:
+      std::cerr << "sendto: " << strerror(errno) << std::endl;
+      return NETWORK_ERR_SEND_FATAL;
+    }
+  }
+
+  assert(nwrite == sendbuf_.size());
+  sendbuf_.reset();
+
+  return NETWORK_ERR_OK;
+}
+
 int Client::start_interactive_input() {
   int rv;
   if (stdinfd_ == -1) {
@@ -868,40 +908,33 @@ int Client::stop_interactive_input() {
   return 0;
 }
 
-void Client::handle_error(int liberr) {
+int Client::handle_error(int liberr) {
   if (!conn_ || ngtcp2_conn_closed(conn_)) {
-    return;
+    return 0;
   }
 
-  std::array<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> buf;
+  sendbuf_.reset();
+  assert(sendbuf_.left() >= max_pktlen_);
 
-  auto n = ngtcp2_conn_write_connection_close(conn_, buf.data(), max_pktlen_,
-                                              infer_quic_error_code(liberr));
+  auto n = ngtcp2_conn_write_connection_close(
+      conn_, sendbuf_.wpos(), max_pktlen_, infer_quic_error_code(liberr));
   if (n < 0) {
     std::cerr << "ngtcp2_conn_write_connection_close: " << ngtcp2_strerror(n)
               << std::endl;
-    return;
+    return -1;
   }
 
-  if (debug::packet_lost(config.tx_loss_prob)) {
-    std::cerr << "** Simulated outgoing packet loss **" << std::endl;
-    return;
-  }
+  sendbuf_.push(n);
 
-  auto nwrite =
-      sendto(fd_, buf.data(), n, 0, &remote_addr_.su.sa, remote_addr_.len);
-  if (nwrite == -1) {
-    std::cerr << "sendto: " << strerror(errno) << std::endl;
-    return;
-  }
+  return send_packet();
 }
 
 namespace {
 void remove_tx_stream_data(std::deque<Buffer> &d, size_t &idx,
                            uint64_t &tx_offset, uint64_t offset) {
-  for (; !d.empty() && tx_offset + d.front().buf.size() <= offset;) {
+  for (; !d.empty() && tx_offset + d.front().tail <= offset;) {
     --idx;
-    tx_offset += d.front().buf.size();
+    tx_offset += d.front().tail;
     d.pop_front();
   }
 }
--- a/examples/network.h
+++ b/examples/network.h
@@ -44,6 +44,12 @@
 
 namespace ngtcp2 {
 
+enum {
+  NETWORK_ERR_OK = 0,
+  NETWORK_ERR_SEND_FATAL = -10,
+  NETWORK_ERR_SEND_NON_FATAL = -11
+} network_error;
+
 union sockaddr_union {
   sockaddr_storage storage;
   sockaddr sa;
--- a/examples/server.cc
+++ b/examples/server.cc
@@ -62,9 +62,9 @@ constexpr size_t MAX_BYTES_IN_FLIGHT = 1460 * 10;
 } // namespace
 
 Buffer::Buffer(const uint8_t *data, size_t datalen)
-    : buf{data, data + datalen}, pos(std::begin(buf)) {}
-
-Buffer::Buffer() : pos(std::begin(buf)) {}
+    : buf{data, data + datalen}, head(0), tail(datalen) {}
+Buffer::Buffer(size_t datalen) : buf(datalen), head(0), tail(0) {}
+Buffer::Buffer() : head(0), tail(0) {}
 
 namespace {
 int bio_write(BIO *b, const char *buf, int len) {
@@ -330,9 +330,8 @@ int Stream::buffer_file() {
   }
 
   for (;;) {
-    auto v = Buffer{};
-    v.buf.resize(16_k);
-    auto nread = read(fd, v.buf.data(), v.buf.size());
+    auto v = Buffer{16_k};
+    auto nread = read(fd, v.wpos(), v.left());
     if (nread == -1) {
       return -1;
     }
@@ -343,9 +342,8 @@ int Stream::buffer_file() {
       fd = -1;
       break;
     }
-    v.buf.resize(nread);
-    v.pos = std::begin(v.buf);
-    streambuf_bytes += v.buf.size();
+    v.push_resize(nread);
+    streambuf_bytes += v.size();
     streambuf.emplace_back(std::move(v));
   }
 
@@ -377,15 +375,14 @@ void Stream::send_status_response(unsigned int status_code,
     hdr += "\r\n";
   }
 
-  auto v = Buffer{};
-  v.buf.resize(hdr.size() + (htp.method == HTTP_HEAD ? 0 : body.size()));
+  auto v = Buffer{hdr.size() + ((htp.method == HTTP_HEAD) ? 0 : body.size())};
   auto p = std::begin(v.buf);
   p = std::copy(std::begin(hdr), std::end(hdr), p);
   if (htp.method != HTTP_HEAD) {
     p = std::copy(std::begin(body), std::end(body), p);
   }
-  v.pos = std::begin(v.buf);
-  streambuf_bytes += v.buf.size();
+  v.push(std::distance(std::begin(v.buf), p));
+  streambuf_bytes += v.size();
   streambuf.emplace_back(std::move(v));
   should_send_fin = true;
   resp_state = RESP_COMPLETED;
@@ -439,12 +436,11 @@ int Stream::start_response() {
     }
     hdr += "\r\n";
 
-    auto v = Buffer{};
-    v.buf.resize(hdr.size());
+    auto v = Buffer{hdr.size()};
     auto p = std::begin(v.buf);
     p = std::copy(std::begin(hdr), std::end(hdr), p);
-    v.pos = std::begin(v.buf);
-    streambuf_bytes += v.buf.size();
+    v.push(std::distance(std::begin(v.buf), p));
+    streambuf_bytes += v.size();
     streambuf.emplace_back(std::move(v));
   }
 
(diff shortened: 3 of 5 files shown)


examples/client.cc
```diff
@@ -795,6 +800,36 @@ ssize_t Client::decrypt_data(uint8_t *dest, size_t destlen,
...
+
+  int eintr_retries = 5;
+  ssize_t nwrite = 0;
+
+retry_send:
+  nwrite =
+    sendto(fd_, data, len, 0, &remote_addr_.su.sa, remote_addr_.len);
+  if (nwrite == -1) {
+    if ((errno == EINTR) && (--eintr_retries > 0)) {
+      goto retry_send;
```
tatsuhiro-t: Personally, I like simple while loop instead of goto.

examples/client.cc
```diff
@@ -795,6 +800,36 @@ ssize_t Client::decrypt_data(uint8_t *dest, size_t destlen,
...
+    }
+    if ((errno == EAGAIN) ||
+        (errno == EWOULDBLOCK) ||
+        (errno == EINTR) ||
+        (errno == 0)) {
```
tatsuhiro-t: switch ... case would be easier to follow?

examples/network.h
```diff
@@ -44,6 +44,12 @@
 
 namespace ngtcp2 {
 
+typedef enum {
```
tatsuhiro-t: In C++, we can just write enum network_error {...}.

examples/server.cc
```diff
@@ -898,9 +898,17 @@ int Handler::on_read(uint8_t *data, size_t datalen) {
 
 int Handler::on_write() {
   std::array<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> buf;
-
   assert(buf.size() >= max_pktlen_);
 
+  while (!sendbuf_.empty()) {
+    auto &v = *std::begin(sendbuf_);
```
tatsuhiro-t: Perhaps, it would be simpler:

```cpp
auto &v = sendbuf_.front();
```

examples/client.cc
```diff
@@ -564,11 +566,21 @@ int Client::on_write() {
   std::array<uint8_t, NGTCP2_MAX_PKTLEN_IPV4> buf;
   assert(buf.size() >= max_pktlen_);
 
+  while (!sendbuf_.empty()) {
+    auto &v = *std::begin(sendbuf_);
```
tatsuhiro-t: Perhaps, it would be simpler to write:

```cpp
auto &v = sendbuf_.front();
```

examples/server.cc
```diff
@@ -1168,7 +1166,16 @@ void Handler::on_stream_close(uint32_t stream_id) {
 }
 
 namespace {
-void swritecb(struct ev_loop *loop, ev_io *w, int revents) {}
+void swritecb(struct ev_loop *loop, ev_io *w, int revents) {
+  auto s = static_cast<Server *>(w->data);
+
+  auto rv = s->on_write();
+  if (rv == NETWORK_ERR_SEND_NON_FATAL) {
+    ev_io_start(loop, w);
```
tatsuhiro-t: It looks like there is no ev_io_stop other than disconnect().  This might cause busy loop.

examples/server.cc
```diff
@@ -1228,6 +1235,25 @@ int Server::init(int fd) {
   return 0;
 }
 
+int Server::on_write() {
+  int rv = NETWORK_ERR_OK;
+
+  for (auto it = std::begin(handlers_);
+       it != std::end(handlers_); ++it) {
+    Handler *h = it->second.get();
```
tatsuhiro-t: Just let auto to figure out the type: `auto h = it->second.get()`

examples/server.cc
```diff
@@ -1228,6 +1235,25 @@ int Server::init(int fd) {
...
+
+    auto hrv = h->on_write();
+    if (hrv == NETWORK_ERR_SEND_NON_FATAL) {
+      rv = NETWORK_ERR_SEND_NON_FATAL;
+    }
+    if (hrv == NETWORK_ERR_SEND_FATAL) {
+      return NETWORK_ERR_SEND_FATAL;
```
tatsuhiro-t: Personally, I think that switch ... case is more readable here.

tatsuhiro-t:
Thank you for PR!  I think the PR is going to good direction.
I'm wondering that using std::deque for sendbuf_ is a bit overkill because we only buffer up at most 1 packet.
Perhaps, allocating std::array per Handler or Client, and the length of written bytes would be simpler because it eliminates the need for queue, and memory copy (sendbuf_.emplace_back).

As for adding another call back to send packet, I think the current API does not need to change.
I guess that primary reason of adding another callback is setting more precise timestamp in ngtcp2_rtb_entry.
Even if we write a packet to the kernel buffer, it does not mean the packet is sent immediately, but we already set timestamp in ngtcp2_rtb_entry.  So I'm not sure how much gain we get if we do care about timestamp skew for only the packet which cannot be written to kernel buffer.  Personally, I'd like to avoid callback if possible.  It creates possibly another layer of indirection.

nickfajones:
> Perhaps, allocating std::array per Handler or Client, and the length of written bytes would be simpler because it eliminates the need for queue, and memory copy (sendbuf_.emplace_back).

That's a good point, according to my own implementation, sending packets to the network will stop as soon as the first packet fails, so I will only need to retain that one packet, so I agree something like using a single std::array per Handler or Client will be ok, and this array can be used as the write buffer for all packets in the connection.  I will use this approach when I update this PR.

> So I'm not sure how much gain we get if we do care about timestamp skew for only the packet which cannot be written to kernel buffer.

I'd like to discuss this quickly for a future PR.  The situation I had in mind is this:
- Let's say an application is writing 3000 bytes to the transport layer (QUIC, ngtcp2).
- Due to MTU and implementation details, this will get split into three packets, each with a STREAM frame of 1000 bytes.
- The `ngtcp2_conn_write_stream` function is called for the first time, this constructs and encodes the first packet and creates an rtb.  The new function: `send_packet` is called and it returns `NETWORK_ERR_OK (0)`
- The `ngtcp2_conn_write_stream` function is called for the second time, constructing and encoding the second packet and creating an rtb, `send packet` is called but this time it returns `NETWORK_ERR_SEND_NON_FATAL`, the second packet is stored and the socket fd is watched for write ability
- After the socket fd is write able, the stored packet is written to network using `send_packet`, and it return  `ngtcp2_conn_write_stream` function is called `NETWORK_ERR_OK (0)`
- The `ngtcp2_conn_write_stream` function is called for the third time, constructing, encoding the third packet and creating an rtb, `send_packet` is called and returns `NETWORK_ERR_OK (0)`

So in this case, the second and third packets are send next to eachother, but the rtb timestamps are quite different.  This is the skew you were talking about, and I worry that it is important to avoid this skew. 
 However, I can't give you a scientific explanation why, it's just a feeling.  Anyhow, for this PR, the current situation will remain unchanged: the rtb entry will be created when the packet is encoded, not when it is written.

>Personally, I'd like to avoid callback if possible. It creates possibly another layer of indirection.

I was thinking of a callback as part of your `ngtcp2_conn_callbacks` structure: adding it in like so:
```
typedef struct {
   ngtcp2_send_client_initial send_client_initial;
   ngtcp2_send_client_cleartext send_client_cleartext;
   ngtcp2_send_server_cleartext send_server_cleartext;
   ngtcp2_recv_handshake_data recv_handshake_data;
+  ngtcp2_write_pkt write_pkt;
   ngtcp2_send_pkt send_pkt;
   ngtcp2_send_frame send_frame;
   ngtcp2_recv_pkt recv_pkt;
   ngtcp2_recv_frame recv_frame;
   ngtcp2_handshake_completed handshake_completed;
   ngtcp2_recv_version_negotiation recv_version_negotiation;
   ngtcp2_encrypt encrypt;
   ngtcp2_decrypt decrypt;
   ngtcp2_recv_stream_data recv_stream_data;
   ngtcp2_acked_stream_data_offset acked_stream_data_offset;
   ngtcp2_stream_close stream_close;
 } ngtcp2_conn_callbacks;
```
Alternatively, we could change the nature of your `send_pkt` to do this, at first I thought this was the purpose of `send_pkt`.

Right now, it may seem that using a callback for writing network data may seem pointless, it is right now but only because the Client and Server(Handle) classes are doing a lot of work: for example the `on_write` and `on_write_stream` functions.  I feel that these functions do a lot of "transport layer" work and these functions may be moved into the core library in future.  By performing the network write in a callback, the "network" side of the application can use return values (like `NETWORK_ERR_SEND_NON_FATAL`) to influence the code path inside the core library, and the library can perform different actions before and after the callback depending on the return value, which is exactly the purpose of the `ngtcp2_conn_callbacks`

Anyhow, I am now discussing a PR that doesn't exist yet.  Tomorrow I will return to this PR and address the comments that you made.

Thanks!

examples/server.cc
nickfajones: I agree with all of the above comments and will address them soon.  Thanks.

tatsuhiro-t:
Properly pacing packet transmission and congestion control may ease the situation a bit.  I haven't implemented it, and actually I haven't read the whole doc, so it is my guess.  Ultimately, it would be nice if kernel offers the information which tells application that how much UDP buffer space left.  In TCP, we use struct tcp_info, and TCP_NOTSENT_LOWAT to avoid excessive buffering in kernel space.

As for client/server complexity, yes they are a bit long.  It is mostly because we have to wire up event handlers, and setup cryptographic context.  For on_write/on_stream_write, they might relate to congestion control.  At the moment, I have an idea to completely separate loss recovery/congestion controller from ngtcp2_conn API, so that it can be used separately.  I haven't decided yet, and as you know, we haven't implemented it either (at the moment, simply limit the number of in flight packet because kernel drops packet if they are written too aggressively).  on_write/on_stream_write is not much complex to me at the moment, and while I think it is good idea to make the example slim, I'd like to keep the current design for a while.

nickfajones:
Hi Tatsuhiro, an update to this PR, addressing the issues you pointed out previously.

I really like your idea about moving all Loss Recovery and Congestion Control away from the QUIC protocol modeller.  I have some ideas on this that I will describe in another issue report.

Thanks!

tatsuhiro-t:
Thank you.
Somehow this PR breaks HTTP/0.9 GET from client.  The short response body is OK.  But long one (e.g., 680KiB) is not sent all.

tatsuhiro-t:
Buffer::left() originally returns the number of bytes left in the buffer to read.
Now it returns the number of bytes we can write it.
Most of the new code uses the latter semantics, but some of original code still uses former.

tatsuhiro-t:
I managed to get the code work.  The patch is attached below.  Note that the diff is a bit large because I run clang-format.

The changes I made is:

* Buffer::left() vs Buffer::size()
* Examine return value of on_write()
* Separate on_read() and on_write()
* Remove Handler from Server on error.

I'm not sure server should shutdown all connections on fatal error.  But BIND probably knows this well.

We should move signint_watcher to Server so that we can stop it on Server::disconnect().  This could be in later PR.

```patch
diff --git a/examples/client.cc b/examples/client.cc
index c12c98e..04cea7f 100644
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -61,8 +61,7 @@ constexpr size_t MAX_BYTES_IN_FLIGHT = 1460 * 10;
 
 Buffer::Buffer(const uint8_t *data, size_t datalen)
     : buf{data, data + datalen}, head(0), tail(datalen) {}
-Buffer::Buffer(size_t datalen)
-    : buf(datalen), head(0), tail(0) {}
+Buffer::Buffer(size_t datalen) : buf(datalen), head(0), tail(0) {}
 Buffer::Buffer() : head(0), tail(0) {}
 
 namespace {
@@ -168,6 +167,11 @@ void readcb(struct ev_loop *loop, ev_io *w, int revents) {
 
   if (c->on_read() != 0) {
     c->disconnect();
+    return;
+  }
+  auto rv = c->on_write();
+  if (rv == NETWORK_ERR_SEND_FATAL) {
+    c->disconnect();
   }
 }
 } // namespace
@@ -572,7 +576,7 @@ int Client::on_read() {
 
   ev_timer_again(loop_, &timer_);
 
-  return on_write();
+  return 0;
 }
 
 int Client::on_write() {
@@ -590,7 +594,7 @@ int Client::on_write() {
     if (rv != 0) {
       return rv;
     }
-    if (v.left() > 0) {
+    if (v.size() > 0) {
       break;
     }
     ++streambuf_idx_;
@@ -648,10 +652,9 @@ int Client::on_write_stream(uint32_t stream_id, uint8_t fin, Buffer &data) {
       break;
     }
 
-    auto n = ngtcp2_conn_write_stream(conn_,
-                                      sendbuf_.wpos(), max_pktlen_, &ndatalen,
-                                      stream_id, fin, data.rpos(), data.size(),
-                                      util::timestamp());
+    auto n = ngtcp2_conn_write_stream(conn_, sendbuf_.wpos(), max_pktlen_,
+                                      &ndatalen, stream_id, fin, data.rpos(),
+                                      data.size(), util::timestamp());
     if (n < 0) {
       if (n == NGTCP2_ERR_STREAM_DATA_BLOCKED) {
         return 0;
@@ -671,7 +674,7 @@ int Client::on_write_stream(uint32_t stream_id, uint8_t fin, Buffer &data) {
       return rv;
     }
 
-    if (data.left() == 0) {
+    if (data.size() == 0) {
       break;
     }
   }
@@ -819,9 +822,8 @@ int Client::send_packet() {
   ssize_t nwrite = 0;
 
   do {
-    nwrite =
-      sendto(fd_, sendbuf_.rpos(), sendbuf_.size(), 0,
-             &remote_addr_.su.sa, remote_addr_.len);
+    nwrite = sendto(fd_, sendbuf_.rpos(), sendbuf_.size(), 0,
+                    &remote_addr_.su.sa, remote_addr_.len);
   } while ((nwrite == -1) && (errno == EINTR) && (eintr_retries-- > 0));
 
   if (nwrite == -1) {
@@ -910,9 +912,8 @@ int Client::handle_error(int liberr) {
   sendbuf_.reset();
   assert(sendbuf_.left() >= max_pktlen_);
 
-  auto n = ngtcp2_conn_write_connection_close(conn_,
-                                              sendbuf_.wpos(), max_pktlen_,
-                                              infer_quic_error_code(liberr));
+  auto n = ngtcp2_conn_write_connection_close(
+      conn_, sendbuf_.wpos(), max_pktlen_, infer_quic_error_code(liberr));
   if (n < 0) {
     std::cerr << "ngtcp2_conn_write_connection_close: " << ngtcp2_strerror(n)
               << std::endl;
@@ -927,9 +928,9 @@ int Client::handle_error(int liberr) {
 namespace {
 void remove_tx_stream_data(std::deque<Buffer> &d, size_t &idx,
                            uint64_t &tx_offset, uint64_t offset) {
-  for (; !d.empty() && tx_offset + d.front().buf.size() <= offset;) {
+  for (; !d.empty() && tx_offset + d.front().tail <= offset;) {
     --idx;
-    tx_offset += d.front().buf.size();
+    tx_offset += d.front().tail;
     d.pop_front();
   }
 }
diff --git a/examples/client.h b/examples/client.h
index 0a0e409..516854d 100644
--- a/examples/client.h
+++ b/examples/client.h
@@ -64,7 +64,7 @@ struct Buffer {
 
   size_t size() const { return tail - head; }
   size_t left() const { return buf.size() - tail; }
-  uint8_t * const wpos() { return &buf[tail]; }
+  uint8_t *const wpos() { return &buf[tail]; }
   const uint8_t *rpos() const { return &buf[head]; }
   void seek(size_t len) { head += len; }
   void push(size_t len) { tail += len; }
diff --git a/examples/server.cc b/examples/server.cc
index 8b10006..3832702 100644
--- a/examples/server.cc
+++ b/examples/server.cc
@@ -63,8 +63,7 @@ constexpr size_t MAX_BYTES_IN_FLIGHT = 1460 * 10;
 
 Buffer::Buffer(const uint8_t *data, size_t datalen)
     : buf{data, data + datalen}, head(0), tail(datalen) {}
-Buffer::Buffer(size_t datalen)
-    : buf(datalen), head(0), tail(0) {}
+Buffer::Buffer(size_t datalen) : buf(datalen), head(0), tail(0) {}
 Buffer::Buffer() : head(0), tail(0) {}
 
 namespace {
@@ -376,8 +375,7 @@ void Stream::send_status_response(unsigned int status_code,
     hdr += "\r\n";
   }
 
-  auto v = Buffer{
-    hdr.size() + ((htp.method == HTTP_HEAD) ? 0 : body.size())};
+  auto v = Buffer{hdr.size() + ((htp.method == HTTP_HEAD) ? 0 : body.size())};
   auto p = std::begin(v.buf);
   p = std::copy(std::begin(hdr), std::end(hdr), p);
   if (htp.method != HTTP_HEAD) {
@@ -480,11 +478,20 @@ void timeoutcb(struct ev_loop *loop, ev_timer *w, int revents) {
 
 namespace {
 void retransmitcb(struct ev_loop *loop, ev_timer *w, int revents) {
+  int rv;
+
   auto h = static_cast<Handler *>(w->data);
+  auto s = h->server();
 
-  if (h->on_write() != 0) {
-    auto server = h->server();
-    server->remove(h);
+  rv = h->on_write();
+  switch (rv) {
+  case 0:
+    break;
+  case NETWORK_ERR_SEND_NON_FATAL:
+    s->start_wev();
+    break;
+  default:
+    s->remove(h);
   }
 }
 } // namespace
@@ -899,18 +906,12 @@ int Handler::on_read(uint8_t *data, size_t datalen) {
 
   ev_timer_again(loop_, &timer_);
 
-  for (auto &p : streams_) {
-    auto &stream = p.second;
-    rv = on_write_stream(*stream);
-    if (rv != 0) {
-      return -1;
-    }
-  }
-
-  return on_write();
+  return 0;
 }
 
 int Handler::on_write() {
+  int rv;
+
   if (sendbuf_.size() > 0) {
     auto rv = server_->send_packet(remote_addr_, sendbuf_);
     if (rv != NETWORK_ERR_OK) {
@@ -920,6 +921,14 @@ int Handler::on_write() {
 
   assert(sendbuf_.left() >= max_pktlen_);
 
+  for (auto &p : streams_) {
+    auto &stream = p.second;
+    rv = on_write_stream(*stream);
+    if (rv != 0) {
+      return rv;
+    }
+  }
+
   for (;;) {
     ssize_t n;
     if (ngtcp2_conn_bytes_in_flight(conn_) < MAX_BYTES_IN_FLIGHT) {
@@ -982,7 +991,7 @@ int Handler::on_write_stream(Stream &stream) {
     if (rv != 0) {
       return rv;
     }
-    if (v.left() > 0) {
+    if (v.size() > 0) {
       break;
     }
     ++stream.streambuf_idx;
@@ -998,11 +1007,9 @@ int Handler::write_stream_data(Stream &stream, int fin, Buffer &data) {
   size_t ndatalen;
 
   for (;;) {
-    auto n = ngtcp2_conn_write_stream(conn_,
-                                      sendbuf_.wpos(), max_pktlen_, &ndatalen,
-                                      stream.stream_id, fin,
-                                      data.rpos(), data.size(),
-                                      util::timestamp());
+    auto n = ngtcp2_conn_write_stream(
+        conn_, sendbuf_.wpos(), max_pktlen_, &ndatalen, stream.stream_id, fin,
+        data.rpos(), data.size(), util::timestamp());
     if (n < 0) {
       if (n == NGTCP2_ERR_STREAM_DATA_BLOCKED) {
         return 0;
@@ -1022,7 +1029,7 @@ int Handler::write_stream_data(Stream &stream, int fin, Buffer &data) {
       return rv;
     }
 
-    if (data.left() == 0) {
+    if (data.size() == 0) {
       break;
     }
   }
@@ -1038,9 +1045,8 @@ int Handler::handle_error(int liberr) {
   sendbuf_.reset();
   assert(sendbuf_.left() >= max_pktlen_);
 
-  auto n = ngtcp2_conn_write_connection_close(conn_,
-                                              sendbuf_.wpos(), max_pktlen_,
-                                              infer_quic_error_code(liberr));
+  auto n = ngtcp2_conn_write_connection_close(
+      conn_, sendbuf_.wpos(), max_pktlen_, infer_quic_error_code(liberr));
   if (n < 0) {
     std::cerr << "ngtcp2_conn_write_connection_close: " << ngtcp2_strerror(n)
               << std::endl;
@@ -1114,10 +1120,11 @@ namespace {
 size_t remove_tx_stream_data(std::deque<Buffer> &d, size_t &idx,
                              uint64_t &tx_offset, uint64_t offset) {
   size_t len = 0;
-  for (; !d.empty() && tx_offset + d.front().buf.size() <= offset;) {
+  for (; !d.empty() && tx_offset + d.front().tail <= offset;) {
     --idx;
-    len += d.front().buf.size();
-    tx_offset += d.front().buf.size();
+    auto &v = d.front();
+    len += v.tail;
+    tx_offset += v.tail;
     d.pop_front();
   }
   return len;
@@ -1178,10 +1185,12 @@ void swritecb(struct ev_loop *loop, ev_io *w, int revents) {
   auto s = static_cast<Server *>(w->data);
 
   auto rv = s->on_write();
-  if (rv == NETWORK_ERR_SEND_NON_FATAL) {
-    ev_io_start(loop, w);
-  } else {
-    s->disconnect();
+  if (rv != 0) {
+    if (rv == NETWORK_ERR_SEND_NON_FATAL) {
+      s->start_wev();
+    } else {
+      s->disconnect();
+    }
   }
 }
 } // namespace
@@ -1244,22 +1253,22 @@ int Server::init(int fd) {
 }
 
 int Server::on_write() {
-  int rv = NETWORK_ERR_OK;
-
-  for (auto it = std::begin(handlers_);
-       it != std::end(handlers_); ++it) {
+  for (auto it = std::begin(handlers_); it != std::end(handlers_);) {
     auto h = it->second.get();
-    auto hrv = h->on_write();
-    switch (hrv) {
+    auto rv = h->on_write();
+    switch (rv) {
+    case 0:
+      ++it;
+      continue;
     case NETWORK_ERR_SEND_NON_FATAL:
-      rv = NETWORK_ERR_SEND_NON_FATAL;
-      break;
+      return NETWORK_ERR_SEND_NON_FATAL;
     case NETWORK_ERR_SEND_FATAL:
       return NETWORK_ERR_SEND_FATAL;
     }
+    it = handlers_.erase(it);
   }
 
-  return rv;
+  return NETWORK_ERR_OK;
 }
 
 int Server::on_read() {
@@ -1321,6 +1330,16 @@ int Server::on_read() {
     if (h->on_read(buf.data(), nread) != 0) {
       return 0;
     }
+    rv = h->on_write();
+    switch (rv) {
+    case 0:
+      break;
+    case NETWORK_ERR_SEND_NON_FATAL:
+      start_wev();
+      break;
+    default:
+      return 0;
+    }
 
     conn_id = h->conn_id();
     handlers_.emplace(conn_id, std::move(h));
@@ -1330,6 +1349,17 @@ int Server::on_read() {
   auto h = (*handler_it).second.get();
   if (h->on_read(buf.data(), nread) != 0) {
     handlers_.erase(conn_id);
+    return 0;
+  }
+  rv = h->on_write();
+  switch (rv) {
+  case 0:
+    break;
+  case NETWORK_ERR_SEND_NON_FATAL:
+    start_wev();
+    break;
+  default:
+    handlers_.erase(conn_id);
   }
 
   return 0;
@@ -1417,7 +1447,7 @@ int Server::send_version_negotiation(const ngtcp2_pkt_hd *chd,
   return 0;
 }
 
-int Server::send_packet(Address& remote_addr, Buffer& buf) {
+int Server::send_packet(Address &remote_addr, Buffer &buf) {
   if (debug::packet_lost(config.tx_loss_prob)) {
     std::cerr << "** Simulated outgoing packet loss **" << std::endl;
     return NETWORK_ERR_OK;
@@ -1427,9 +1457,8 @@ int Server::send_packet(Address& remote_addr, Buffer& buf) {
   ssize_t nwrite = 0;
 
   do {
-    nwrite =
-      sendto(fd_, buf.rpos(), buf.size(), 0,
-             &remote_addr.su.sa, remote_addr.len);
+    nwrite = sendto(fd_, buf.rpos(), buf.size(), 0, &remote_addr.su.sa,
+                    remote_addr.len);
   } while ((nwrite == -1) && (errno == EINTR) && (eintr_retries-- > 0));
 
   if (nwrite == -1) {
@@ -1452,6 +1481,8 @@ int Server::send_packet(Address& remote_addr, Buffer& buf) {
 
 void Server::remove(const Handler *h) { handlers_.erase(h->conn_id()); }
 
+void Server::start_wev() { ev_io_start(loop_, &wev_); }
+
 namespace {
 int alpn_select_proto_cb(SSL *ssl, const unsigned char **out,
                          unsigned char *outlen, const unsigned char *in,
diff --git a/examples/server.h b/examples/server.h
index 43f6587..503fb91 100644
--- a/examples/server.h
+++ b/examples/server.h
@@ -69,7 +69,7 @@ struct Buffer {
 
   size_t size() const { return tail - head; }
   size_t left() const { return buf.size() - tail; }
-  uint8_t * const wpos() { return &buf[tail]; }
+  uint8_t *const wpos() { return &buf[tail]; }
   const uint8_t *rpos() const { return &buf[head]; }
   void seek(size_t len) { head += len; }
   void push(size_t len) { tail += len; }
@@ -77,9 +77,7 @@ struct Buffer {
     push(len);
     buf.resize(tail);
   }
-  void resize(size_t size) {
-    buf.resize(size);
-  }
+  void resize(size_t size) { buf.resize(size); }
   void reset() {
     head = 0;
     tail = 0;
@@ -223,8 +221,9 @@ public:
   int on_read();
   int send_version_negotiation(const ngtcp2_pkt_hd *hd, const sockaddr *sa,
                                socklen_t salen);
-  int send_packet(Address& remote_addr, Buffer& buf);
+  int send_packet(Address &remote_addr, Buffer &buf);
   void remove(const Handler *h);
+  void start_wev();
 
 private:
   std::map<uint64_t, std::unique_ptr<Handler>> handlers_;
```

nickfajones:
It was poor judgement on my part to include the changes to the Buffer class in the PR.  I'm sorry.  I will re-submit, just focusing on the changes to on_write as originally intended + addressing your original feedback + adding the changes in your patch above such as: checking on_read return value, start_wev and remove handler from the server on error.

I'll stay away from changing the Buffer class and leave it as a future PR as it has caused confusion for both of us.

Very sorry.

tatsuhiro-t:
@nickfajones If you are OK, I'd like to merge this PR with my patch.  I plan to change examples/client to support multiple streams, and doing it probably makes a lot of conflict.

nickfajones:
Hi Tatsuhiro, I have updated the PR with your patch and no additional changes.  You may pull this PR or simply integrate your own work from which you generated the patch originally.

Some notes:

- There is still some inconsistent usage of the 'new' Buffer: in the Client and Server: `remove_tx_stream_data` functions where you access the `.tail` member variable directly, I suggest that the `size()` function be used as this takes into account not only the value of `.tail` but also the value of the `.head` member variable.  I feel that the code is not entirely correct now as it simply stops freeing buffers if the acked data offset finishes inside a Buffer, in this case I feel that the `seek` method should be called on the Buffer, and the Buffer be removed later when a subsequent ack is received.  I will wait until you have finished your feature to add multiple stream support to example/client, and revisit this code if you don't change it yourself

- Since you will now be supporting multiple streams in the client (and server) you might want to revisit the idea of having a `std::deque` of Buffer objects as I did in my original PR, as each stream may need to queue packets

- The reference I made to BIND was merely for checking the return value of `sendto`, including the strategy of trying again if `EINTR` is the cause of the error, I didn't do any research on how BIND handles failure of the socket.  I agree that shutting down the connection is not a good thing to do if there is a socket error, I think there should be some kind of strategy for re-binding and re-opening the UDP socket.  The delay caused by this can be handled by the `ngtcp2` core library in the same way that it would handle transport level delays.  Again, I may look to BIND or other UDP services for inspiration to see how they handle socket errors

tatsuhiro-t:
> There is still some inconsistent usage of the 'new' Buffer: in the Client and Server: ...  I feel that the code is not entirely correct now as it simply stops freeing buffers if the acked data offset finishes inside a Buffer

I think the current code is fine.  We always start writing data into buffer from index 0.  So .tail shows how much data is written in the buffer.  remove_tx_stream_data simply checks that entire buffer is acked, or not, and if so, remove the buffer.

> Since you will now be supporting multiple streams in the client (and server) you might want to revisit the ...

The server is already multiple stream capable, and it only has a single buffer per handler (single QUIC connection).  The client still has one connection, so we only need a single buffer.  Notice that packet has sequential packet number per a connection.

>  I agree that shutting down the connection is not a good thing to do if there is a socket error, I think there should be some kind of strategy for re-binding and re-opening the UDP socket.

I think it would be safer now to just remove the handler instead of shutting down entire server when we got socket error.

PR #43, opened 2017-09-28
prefer connect() + send() than sendto()

zizhong:
On Mac, sendto() may have a stricter implementation.
It can return `Socket is already connected` error.

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -473,6 +473,12 @@ int Client::init(int fd, const Address &remote_addr, const char *addr,
   fd_ = fd;
   datafd_ = datafd;
 
+
+  if (-1 == connect(fd_, &remote_addr_.su.sa, remote_addr_.len)) {
+    std::cerr << "connect: " << strerror(errno) << std::endl;
+    return -1;
+  }
+
   ssl_ = SSL_new(ssl_ctx_);
   auto bio = BIO_new(create_bio_method());
   BIO_set_data(bio, this);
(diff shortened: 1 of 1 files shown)


examples/client.cc
```diff
@@ -866,12 +866,17 @@ int Client::send_packet() {
     return NETWORK_ERR_OK;
   }
 
+  if (-1 == connect(fd_, &remote_addr_.su.sa, remote_addr_.len)) {
+    sendbuf_.reset();
+    std::cerr << "connect: " << strerror(errno) << std::endl;
+    return NETWORK_ERR_OK;
+  }
```
tatsuhiro-t: I think connect(2) can be done just only once in run().  Does mac os require to call connect(2) every time a packet is sent?

zizhong:
Yes, you're right. fixed. Thanks for pointing it out.

examples/client.cc
```diff
@@ -473,6 +473,12 @@ int Client::init(int fd, const Address &remote_addr, const char *addr,
   fd_ = fd;
   datafd_ = datafd;
 
+
+  if (-1 == connect(fd_, &remote_addr_.su.sa, remote_addr_.len)) {
+    std::cerr << "connect: " << strerror(errno) << std::endl;
+    return NETWORK_ERR_OK;
```
tatsuhiro-t: This should be just `return -1`.

zizhong:
My bad. Fixed.

tatsuhiro-t:
Cherry picked as 4919f221f82b8e59fc1fd062b71f2586ce80e108.  Thank you!

PR #50, opened 2018-01-14
travis: fix rpath

alagoutte:
to avoid ./examples/client: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory


tatsuhiro-t:
Thank you for PR!  After successful Ci build, I'll merge this PR.

alagoutte:
it is now good :)

PR #51, opened 2018-01-15
fix ngtcp2.h: '@function' command should be used in a comment attache…

alagoutte:
…d to a function declaration [-Werror,-Wdocumentation]


tatsuhiro-t:
Thank you for PR.  It looks like it is really a bug.  The thing is that it should be `@functypedef` rather than `@functiontypedef`.

alagoutte:
Updated !

tatsuhiro-t:
Thanks!

PR #53, opened 2018-01-19
Simplify test certificate creation

Lekensteyn:
At the same time, remove the "gen-certificate" target" since it is not
required as part of the build process and not documented elsewhere. This
script could be used for testing "examples/server".
___
It basically reverts #29 and adds a new script on top. @alagoutte I think you already have modifications to .travis.yml that invoke the test programs, is that correct?

FWIW @alagoutte mentioned that the example programs initially failed to run (fixed by #50) with autotools, but when testing with CMake I found no issues.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -24,13 +24,6 @@ SUBDIRS = lib tests third-party examples
 
 ACLOCAL_AMFLAGS = -I m4
 
-CERTIFICATE_PASS=QUIC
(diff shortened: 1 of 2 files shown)


alagoutte:
LGTM (work on my working branch https://github.com/alagoutte/ngtcp2/tree/travis_test_with_peter / https://travis-ci.org/alagoutte/ngtcp2/builds/330782043

tatsuhiro-t:
Thank you for PR.  This looks good.

PR #66, opened 2018-03-09
Fix issues with scan-build (clang)

alagoutte:


tatsuhiro-t:
I think efbde304769c9a6e78712681288e00e1960c29cf is false positive.  Technically the last assignment can be removed, but I'd like to leave it as is because it would be easier to add frames later.

alagoutte:
Patches updated

lib/ngtcp2_conn.c
```diff
@@ -1212,7 +1212,6 @@ static ssize_t conn_write_handshake_pkt(ngtcp2_conn *conn, uint8_t *dest,
     }
 
     *pfrc = frc;
-    pfrc = &frc->next;
```
tatsuhiro-t: I might not be clear, I leave this as is.  Because it is very easy to forget to add this when someone adds new chain after this block later.

alagoutte: Done (remove the patch..)

tatsuhiro-t:
Thank you!  Merged now.

PR #72, opened 2018-04-02
Drain all udp packet

scw00:
It will significant improve the ngtcp2's performance.


tatsuhiro-t:
Thank you.  Merged now.

PR #82, opened 2018-08-09
Add unordered_data flag at application level.

davidjwbbc:
This adds an flag to the ngtcp2_settings to change the behaviour of ngtcp2 when it encounters reordered packets or packets following missing packets. When the flag is set it will return stream data to application immediately without waiting for the missing (or reordered) packets. This allows the application to process the stream data while the missing parts are being waited for. It also reduces the buffering overhead in ngtcp2 as no stream data needs to be buffered.

The patch maintains existing gaps/data records for the stream, but if the unordered_data flag is set then the actual data buffer stored is a NULL pointer and the data structure contains no data part to indicate that the data was not kept and has already been passed onto the application. These blocks can then be skipped when reassembling the stream.

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -503,6 +503,10 @@ typedef struct {
    ngtcp2_conn_server_new. */
 typedef void (*ngtcp2_printf)(void *user_data, const char *format, ...);
 
+typedef enum {
+  NGTCP2_SETTINGS_FLAG_UNORDERED_DATA = 0x1u
+} ngtcp2_settings_flags;
+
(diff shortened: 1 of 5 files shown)


XingkaiLiang:
@davidjwbbc is there any schedule to update the nghq to the latest version following latest ngtcp2 ? Thanks.

LPardue:
Hello @XingkaiLiang. We have an ambition to update nghq toward the end of the year (we are a little ties up on other projects at the moment). We're likely to maintain a focus on the multicast HTTP/QUIC implementation and welcome contribution to the unicast HTTP/QUIC side of things.

blaquezzz (approved):

PR #84, opened 2018-09-16
examples: Support writing new keylog format

Lekensteyn:
Since draft -13, the exporter secret has become useless for decryption.
While the client and server programs log appropriate secrets using the
"-s" option, let's make integration with Wireshark easier by including
the appropriate labels and identifier (client random).

The corresponding Wireshark patch that uses these secrets:
https://code.wireshark.org/review/29677
(tested with ngtcp2 master (draft -13) and draft-14)


tatsuhiro-t:

PR #96, opened 2018-11-30
conn_shutdown_stream_read() does not work correctly

NKTelnet:
@tatsuhiro-t

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -3816,10 +3816,10 @@ static int conn_recv_stream(ngtcp2_conn *conn, const ngtcp2_stream *fr) {
     if (fr_end_offset <= rx_offset) {
       return 0;
     }
+  }
 
-    if (strm->flags & NGTCP2_STRM_FLAG_STOP_SENDING) {
-      return 0;
-    }
+  if (strm->flags & NGTCP2_STRM_FLAG_STOP_SENDING) {
+    return 0;
   }
 
   if (fr->offset <= rx_offset) {
@@ -6089,7 +6089,7 @@ static int conn_shutdown_stream_read(ngtcp2_conn *conn, ngtcp2_strm *strm,
     return 0;
   }
 
-  strm->flags |= NGTCP2_STRM_FLAG_STOP_SENDING;
+  strm->flags |= NGTCP2_STRM_FLAG_STOP_SENDING | NGTCP2_STRM_FLAG_RECV_RST;
   strm->app_error_code = app_error_code;
 
   return conn_stop_sending(conn, strm, app_error_code);


tatsuhiro-t:
Could you explain what this PR fixes?  Specifically:

- what happened
- what you expected to happen
- how to reproduce it

lib/ngtcp2_conn.c
```diff
@@ -3816,10 +3816,10 @@ static int conn_recv_stream(ngtcp2_conn *conn, const ngtcp2_stream *fr) {
     if (fr_end_offset <= rx_offset) {
       return 0;
     }
+  }
 
-    if (strm->flags & NGTCP2_STRM_FLAG_STOP_SENDING) {
-      return 0;
-    }
+  if (strm->flags & NGTCP2_STRM_FLAG_STOP_SENDING) {
```
NKTelnet, line 3821: This change ensures no data can be forwarded to application layer no matter whether the fin flag in the frame is set or not

lib/ngtcp2_conn.c
```diff
@@ -6089,7 +6089,7 @@ static int conn_shutdown_stream_read(ngtcp2_conn *conn, ngtcp2_strm *strm,
     return 0;
   }
 
-  strm->flags |= NGTCP2_STRM_FLAG_STOP_SENDING;
+  strm->flags |= NGTCP2_STRM_FLAG_STOP_SENDING | NGTCP2_STRM_FLAG_RECV_RST;
```
NKTelnet, line 6092: This change ensure the stream can be closed correctly. If NGTCP2_STRM_FLAG_RECV_RST is not set, the stream can not be closed by ngtcp2_conn_close_stream_if_shut_rdwr(). NGTCP2_STRM_FLAG_RECV_RST will be set by conn_recv_rst_stream(), but peer may not send RST_STREAM frame when receiving STOP_SENDING frame because the stream may has been closed by peer. If peer did not send RST_STREAM frame, the NGTCP2_STRM_FLAG_RECV_RST will not be set, and the related stream can not be closed. "ngtcp2_rob_first_gap_offset(&strm->rob) == strm->last_rx_offset" can be always false because of one loss packet and NGTCP2_STRM_FLAG_STOP_SENDING flag.

lib/ngtcp2_conn.c
tatsuhiro-t: I think this particular issue has been fixed in draft-15 branch.

lib/ngtcp2_conn.c
tatsuhiro-t: In QUIC, peer has to close its send stream either STREAM with fin bit set or RST_STREAM.
Even if peer receives STOP_SENDING, it has to do either sending STREAM with fin or RST_STREAM.  If peer does not do this, it is broken.

NKTelnet: If the related stream has been closed by peer, How can peer send RST_STREAM or fin ?

lib/ngtcp2_conn.c
NKTelnet: It is better to move the code to common place ?

tatsuhiro-t: In general yes.  For this particular change, the current code works fine and I don't feel any need for change.

lib/ngtcp2_conn.c
tatsuhiro-t: What is "related" stream?
Bidirectional stream has states for both directions independently.  Only sending peer can close it.
If an endpoint receives RST_STREAM from peer, it has to send STREAM FIN or RST_STREAM to close its sending stream.  Receiving RST_STREAM does not mean both side of bidirectional stream are closed.

PR #97, opened 2018-12-05
ngtcp2_conn_write_stream() returns incorrect error

NKTelnet:
when the congestion window size of a connection is small (lower than 36?), ngtcp2_conn_write_stream() may return ERR_STREAM_DATA_BLOCKED, but the flow control windows are big enough at that time.

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -1390,10 +1390,7 @@ static ssize_t conn_write_handshake_pkts(ngtcp2_conn *conn, uint8_t *dest,
   nwrite = conn_write_handshake_pkt(conn, dest, destlen, NGTCP2_PKT_INITIAL,
                                     early_datalen, ts);
   if (nwrite < 0) {
-    if (nwrite != NGTCP2_ERR_NOBUF) {
-      return nwrite;
-    }
-    return NGTCP2_ERR_NOBUF;
+    return nwrite;
   }
 
   res += nwrite;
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.
There are some issues around NGTCP2_ERR_NOBUF in ngtcp2 and I have a plan to review its usage.
There are similar error code like NGTCP2_ERR_CONGESTION, but it is a bit unclear which error code should be returned.  This issue will be handled in the process.

NKTelnet:
I agree and I think there should not be NGTCP2_ERR_NOBUF error.  It should return NGTCP2_ERR_CONGESTION. All NGTCP2_ERR_NOBUF should be removed and the caller should set the buffer to UDP packet size.

NKTelnet:
By the way, the NGTCP2_MIN_PKTLEN should be set to 37 acording to your code. NGTCP2_MIN_PKTLEN is not the mininal quic packet size, but for congestion control

tatsuhiro-t:
It is good idea to set NGTCP2_MIN_PKTLEN to minimum packet size... the thing is some of the quic header is variable length.
I think NGTCP2_ERR_NOBUF purely indicates that the buffer size application set is too small without congestion control applied.  It is not needed if application sets enough packet size.

tatsuhiro-t:
NGTCP2_ERR_NOMEM is now not surfaced from ngtcp2 API.
I also removed NGTCP2_ERR_CONGESTION.  Now APIs return 0 in the situation that these 2 error codes are returned previously.  I think this is ok because all app should do is keep reading and wait for congestion window to grow.  There is no special treatment for congestion.
NGTCP2_MIN_PKTLEN is also removed.  If we cannot write anything in congestion limited buffer, try ACK only packet without the restriction.
I also fixed couple of bugs and assertion errors.

PR #107, opened 2019-02-08
Fix build failure for --disable-shared

danbev:
Currently, when configuring with --disable-shared the following error is
generated when running `make check`:
```console
make[2]: *** No rule to make target `../lib/.libs/*.o',
needed by `main'.  Stop.
make[2]: *** Waiting for unfinished jobs....
```
This commit adds a conditional for this case.


tatsuhiro-t:
Sorry for responding this late.  It is cherry-picked as 90466b2c9f34e336efc07ad720b99873d2a11cf3
Thank you for PR.

PR #123, opened 2019-06-12
add error argument in `nghttp3_conn_close_stream`

aneesh-joshi:
the current call to  `nghttp3_conn_close_stream` isn't compatible with the definition in nghttp3.


tatsuhiro-t:
Please use draft-22 branch and nghttp3 master for now.

PR #124, opened 2019-06-14
CMake improvements.

DaanDeMeyer:
- Allow specifying shared or static with BUILD_SHARED_LIBS.
- Add include directories as PUBLIC to library targets.

This enables using ngtcp2 with CMake `add_subdirectory`.


tatsuhiro-t:

PR #130, opened 2019-07-14
Remove final newlines from log messages.

DaanDeMeyer:
Adding the final newline is left to the logging implementation.

Logging messages might be stored in a format where newlines are not required. This is also easier to plug into logging implementations that add the final newline themselves.


tatsuhiro-t:
Thank you for PR.  I like this patch.  Could you create PR against draft-22 branch?

DaanDeMeyer:
See #131

PR #137, opened 2019-07-19
Add struct tags for structs used in callbacks.

DaanDeMeyer:
This allows them to be forward declared. Typedefed anonymous structs
cannot be forward declared without a tag.


tatsuhiro-t:

PR #140, opened 2019-08-04
build: Only build examples if nghttp3 is present

falconindy:
This isn't critical to the installation of the core library, so don't force
users to have nghttp3 available.


tatsuhiro-t:
Thank you for PR.  This was cherry-picked (2582cf106cf975c5f07571e56ce3984c7f04bc95) into draft-22 branch with some amends.

PR #143, opened 2019-08-16
fix connection flow control data calculation error

NKTelnet:


lib/ngtcp2_conn.c
```diff
@@ -4893,10 +4893,6 @@ static int conn_recv_stream(ngtcp2_conn *conn, const ngtcp2_stream *fr) {
     }
 
     conn->rx.offset += len;
-
-    if (strm->flags & NGTCP2_STRM_FLAG_STOP_SENDING) {
-      ngtcp2_conn_extend_max_offset(conn, len);
```
NKTelnet, line 4898: ngtcp2_strm_rx_offset(strm) may not equal strm->rx.last_offset here, so we can not extend connection max offset here.

lib/ngtcp2_conn.c
```diff
@@ -5202,7 +5198,6 @@ static int conn_recv_reset_stream(ngtcp2_conn *conn,
   }
 
   conn->rx.offset += datalen;
-  ngtcp2_conn_extend_max_offset(conn, datalen);
```
NKTelnet, line 5205: ngtcp2_strm_rx_offset(strm) may not equal strm->rx.last_offset here, so we can not simply extend connection max offset by strm->rx.last_offset.

tatsuhiro-t:
Could you describe the condition/situation when the flow control calculation error occurs?

NKTelnet:
> Could you describe the condition/situation when the flow control calculation error occurs?

As I have said above, when packets loss happened, ngtcp2_strm_rx_offset(strm) may not equal strm->rx.last_offset. So we can not use strm->rx.last_offset to update connection flow control window. Application calls ngtcp2_conn_extend_max_offset() only for the data before ngtcp2_strm_rx_offset(strm).

tatsuhiro-t:
Regardless of packet loss, when an endpoint sees strm->rx.last_offset, the remote peer committed stream data up to strm->rx.last_offset.   This means that we can extend connection window if we no longer pass any data to app (STOP_SENDING has sent or RESET_STREAM has received).

tatsuhiro-t:
I might have missed the details of buffered data which is not passed to the app.  Let me check.

NKTelnet:
> I might have missed the details of buffered data which is not passed to the app. Let me check.

Not only buffered data, buffered data is another topic. The loss packets will cause this issue

tatsuhiro-t:
The bug I try to confirm is that buffered STREAM data not passed to app is not added back to connection window.  Have you identified any other issue?

NKTelnet:
> The bug I try to confirm is that buffered STREAM data not passed to app is not added back to connection window. Have you identified any other issue?

If you define unordered STREAM data as buffered STREAM data, I think it is ok. My patch will resolve this issue.

Suppose a stream has received data from 0 to 100 and from 200 to 300, the data from 101 to 199 have been lost. The strm->rx.last_offset should be 300 now, but ngtcp2_strm_rx_offset(strm) is 100.   
At this time, the stream received RST with final_size 300, the connection flow control window will not be updated any more, but receiver application only has updated the connection flow control window with length 100. 200 bytes window will be lost.

tatsuhiro-t:
Thanks.  ngtcp2 does not support unordered STREAM.  OK, so the only issue is buffered data is not added back to connection window.  I think it would be nicer to extend connection window immediately to minimize the chance of window exhaustion rather than waiting for stream to close.  Closing stream might be delayed.

NKTelnet:
> Thanks. ngtcp2 does not support unordered STREAM. OK, so the only issue is buffered data is not added back to connection window. I think it would be nicer to extend connection window immediately to minimize the chance of window exhaustion rather than waiting for stream to close. Closing stream might be delayed.

Wait. I think we must have some misunderstanding from each other.

In function conn_recv_stream(ngtcp2_conn *conn, const ngtcp2_stream *fr), the code will call:

  } else if (fr->datacnt) {
    rv = ngtcp2_strm_recv_reordering(strm, fr->data[0].base, fr->data[0].len,
                                     fr->offset);

to buffer unordered data and this will cause ngtcp2_strm_rx_offset(strm) not equal to  strm->rx.last_offset

tatsuhiro-t:
I don't follow.  It is generally ok that ngtcp2_strm_rx_offset != strm->rx.last_offset.  If they must be the same, we don't have to remember both of them.  Could you explain that what the problem is?

NKTelnet:
> I don't follow. It is generally ok that ngtcp2_strm_rx_offset != strm->rx.last_offset. If they must be the same, we don't have to remember both of them. Could you explain that what the problem is?

In your code, you use strm->rx.last_offset to update connection window in the corner cases and application will use ngtcp2_strm_rx_offset to update connection window, this is the root cause of this issue.  

"The bug I try to confirm is that buffered STREAM data not passed to app is not added back to connection window." This is not true, if the stream does not have buffered STREAM data not passed to app, this issue will happen too, unless you call unordered data as buffered data. How do you define "buffered STREAM data" ? Is the data ready to be passed to app or the data in reordering list?

tatsuhiro-t:
The buffered STREAM data is data buffered in ngtcp2_conn and not yet passed to application because of packet reordering or loss.

NKTelnet:
> The buffered STREAM data is data buffered in ngtcp2_conn and not yet passed to application because of packet reordering or loss.

Ok, so I agree. And how will you fix this issue ? It seems you have to add a member to ngtcp2_strm struct to track the ngtcp2_strm_rx_offset when stream is in normal state, and to track the window when stream is in abnormal state.

tatsuhiro-t:
The current plan is extend connection window for "buffered STREAM data" (strm->rx.last_offset - ngtcp2_strm_rx_offset(strm)) when first sending STOP_SENDING or receiving RESET_STREAM.  I think no additional field is necessary.

NKTelnet:
> The current plan is extend connection window for "buffered STREAM data" (strm->rx.last_offset - ngtcp2_strm_rx_offset(strm)) when first sending STOP_SENDING or receiving RESET_STREAM. I think no additional field is necessary.

How do you keep the information that (strm->rx.last_offset - ngtcp2_strm_rx_offset(strm)) has been added to connection window ?

tatsuhiro-t:
We have flags NGTCP2_STRM_FLAG_STOP_SENDING and NGTCP2_STRM_FLAG_RECV_RST.
Using these flags, update connection window just once.

tatsuhiro-t:
Fix has committed via e2126a5db3dbe68cf3bf600e113c45433e20e0f5

PR #151, opened 2019-09-16
Remove chunk comment from ngtcp2_idtr_init

danbev:
This commit removes the comment for a `chunk` argument for
`ngtcp2_idtr_init` as it does not seem to be relative to this function
declaration.


tatsuhiro-t:
Thank you for PR.  Cherry-picked via a7eb50231e975534afe5121fcd38434c03c41052 into master.

PR #152, opened 2019-09-16
Remove duplication deque include from server.h

danbev:


tatsuhiro-t:
Thank you for PR.  Cherry-picked via 4bcf37efd485e23a634d21c9c573285758a3258d into master.

PR #153, opened 2019-09-16
Remove template from HTTPHeader constructor

danbev:
Currently, the HTTPHeader constructor is templated with the types for
the header name and the value, but the actual name and value members are
of std::string type. It is currently not possible to have a different
type for the value for example.

This commit suggests removing the template for the constructor and
making the parameters to to be of type std::string instead.  I'd be
happy to look at a templated version but was not sure if it was worth
the effort as this is "just" in an example.

### DIFF
--- a/examples/server.h
+++ b/examples/server.h
@@ -116,8 +116,8 @@ struct Buffer {
 };
 
 struct HTTPHeader {
-  template <typename T1, typename T2>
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Sorry for late response, this PR was cherry-picked as a854ed6

PR #157, opened 2019-10-09
ngtcp2_crypto: add support for additional handshake err codes

jasnell:
When SSL_dl_handshake() returns either an SSL_ERROR_WANT_CLIENT_HELLO_CB
or SSL_ERROR_WANT_X509_LOOKUP, return specific error codes rather than
-1 so that those conditions can be differentiated and handled
appropriately

### DIFF
--- a/crypto/includes/ngtcp2/ngtcp2_crypto.h
+++ b/crypto/includes/ngtcp2/ngtcp2_crypto.h
@@ -35,6 +35,15 @@ extern "C" {
 #define NGTCP2_CRYPTO_INITIAL_KEYLEN 16
 #define NGTCP2_CRYPTO_INITIAL_IVLEN 12
 
+#if defined(__cplusplus) && __cplusplus >= 201103L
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Thank you for PR.
I understand that differentiating those error code is important for OpenSSL users and ngtcp2 crypto lib should notify this to the caller.  I'm thinking about the best way to incorporate this in a vendor neutral way.  It might be nice to provide a place to store the implementation dependent error code and its header file to describe its own error code.

tatsuhiro-t:
Hmm, it might be sufficient to define common error code and implementation dependent error code space.  The current use of ngtcp2 crypto lib requires application to know the underlying crypto library after all.

tatsuhiro-t:
Cherry-picked as 206eefb32ceffa2e056c2f00dbcb41bfe79061db with some arrangements of error code in a separate file (see 0232fb014b5e450eafa42ab7683e4f80274f25fa)

jasnell:
Thank you!

PR #160, opened 2019-10-11
Make HTTPHeader struct templated

danbev:
Currently, the HTTPHeader constructor is templated with the types for
the header name and the value, but the actual name and value members are
of std::string type. It is currently not possible to have a different
type for the value for example.

This commit suggests making the struct templated to allow for values
other than std::string. I'm not sure if that is the way to go or perhaps
it would be acceptable to only allow std::string values, but wanted to
open this up for feedback.

### DIFF
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -61,7 +61,8 @@ server_SOURCES = server.cc server.h \
 if HAVE_CUNIT
 check_PROGRAMS = examplestest
 examplestest_SOURCES = examplestest.cc \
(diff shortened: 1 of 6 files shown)


tatsuhiro-t:
I think it is just enough to accept std::string for now.
Sorry for not commenting to #153, I think #153 is the way to go.

danbev:
> Sorry for not commenting to #153, I think #153 is the way to go.

No worries at all! Sounds good, I'll update this PR as I opened #153 against the wrong repo. Thanks.

PR #161, opened 2019-10-14
Remove client -i --interactive option from README

danbev:
This commit removes the `-i, --interactive` option for the client example
which was removed in Commit 9ea17af6ef48281d015a73ce46d6689b2255a037 ("h3 client and server").


tatsuhiro-t:

PR #162, opened 2019-10-15
Add missing dash to quiet option comment

danbev:


tatsuhiro-t:

PR #164, opened 2019-10-21
Remove unused ngtcp2_list_remove macro

danbev:


lib/ngtcp2_macro.h
```diff
@@ -31,8 +31,6 @@
 
 #include <stddef.h>
 
-#include <ngtcp2/ngtcp2.h>
-
```
tatsuhiro-t: Could you leave this line and just delete the macro?

danbev: Absolutely, I'll update shortly. Thanks

tatsuhiro-t:
Thank you for prompt update.  Now merged.

PR #165, opened 2019-10-21
Use OpenSSL return values for client_hello_cb

danbev:
This commit updates the `client_hello_cb` to use return values specified
in [SSL_CTX_set_client_hello_cb](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_client_hello_cb.html).


tatsuhiro-t:

PR #166, opened 2019-10-22
Write ACK to Server Initial from client

JunqiWang:
Some implementations of server don't make indications of client having received SHLO and do expect ACK.

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -1720,7 +1720,6 @@ static ssize_t conn_write_handshake_ack_pkt(ngtcp2_conn *conn, uint8_t *dest,
 
   switch (type) {
   case NGTCP2_PKT_INITIAL:
-    assert(conn->server);
     pktns = &conn->in_pktns;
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
I think server should handle this situation without ACK from client.  QUIC is still early stage and those servers should be fixed.
I'll wait for next quic draft to see any changes in this area.

JunqiWang:
I agree server should handle the case. But I think client should also send ACK so that server can start having RTT samples earlier.

tatsuhiro-t:
Getting rtt sample earlier is valid argument.  QUIC wg will change the rules of packtization of Initials in the next draft, so I will revisit this issue when the next draft is published.

JunqiWang:
sounds good. Closing

mjoras:
Can we repen this? The language in the spec is actually now saying that you should acknowledge data at an encryption level before discarding keys:

> An endpoint cannot discard keys for a given encryption level unless it has both received and acknowledged all CRYPTO frames for that encryption level and when all CRYPTO frames for that encryption level have been acknowledged by its peer.

PR #167, opened 2019-10-22
Use char instead of string for \n in client.cc

danbev:
This commit updates `write_transport_params` in client.cc to use a
character `'\n'`, instead of string `"\n"`.


tatsuhiro-t:
Thank you. Merged now.

PR #168, opened 2019-10-22
Remove unused parts of Buffer in example

amirlivneh:
A subsequent PR can further simplify `Buffer` as `begin` is now always equal to `head` and `size()` is always equal to `bufsize()`.


tatsuhiro-t:

PR #169, opened 2019-10-22
Use closure for scid and dcid generation

danbev:
This commit suggests using a closure for the generation of ngtcp2_cids
in Client::init. The motivation for this is to reduce some minor code
duplication.


tatsuhiro-t:

PR #171, opened 2019-10-23
Remove unused ngtcp2_macro include in ngtcp2_qlog

danbev:
This commit removes the include of `ngtcp2_macro.h` as it does not seem to
be used.


tatsuhiro-t:
Thank you.  Merged now.

PR #172, opened 2019-10-23
Update CMakeLists to enable the usage of project Module Path in addition to ngtcp2 one

VALERE91:
Hi  guys,

we had a problem with our CMake modules after the activation of ngtcp2 in out project. This simple fix resolve the issue.

I simply added the base CMAKE_MODULE_PATH to your redefinition to enable your modules first and the project module after.

Thanks


tatsuhiro-t:
Thank you for PR.  Cherry-picked via 5cae63300661a249ac9f1da3d1c4fdb64e171807

PR #173, opened 2019-10-24
Remove redundant Buffer::head

amirlivneh:
After removing `Buffer::seek`, `head` always equals `begin`. A subsequent diff can further simplify Buffer by removing bufsize(), which is now identical to size().


tatsuhiro-t:
Thank you.  Merged now.

PR #178, opened 2019-10-25
Move ngtcp2_range.h into ngtcp2_gaptr.c

danbev:
The header `ngtcp2_gaptr.h` does not use anything from `ngtcp2_range.h` and
this commit suggests moving the include to `ngtcp2_gaptr.c` instead.


tatsuhiro-t:

PR #179, opened 2019-10-25
Move template.h include to client.cc

danbev:
This commit moves the include of `template.h` into `client.cc` as it does
not seem to be used any longer by `client.h`.

Commit a7d6c8c44ce5304c4983d0cc276dd2e6b47f3c1c ("Addngtcp2_acked_stream_data_offset callback")
removed the field streambuf_ which was using the user-defined literal `_k` from template.h:
```c
std::array<uint8_t, 32_k> streambuf_;
```

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -50,6 +50,7 @@
 #include "debug.h"
 #include "util.h"
 #include "shared.h"
+#include "template.h"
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:
Thank you.  Merged now.

PR #180, opened 2019-10-25
Check mmap's return value

amirlivneh:
Previously, when mmap() failed the client example crashed deep inside ngtcp2 on trying to access address 0xffffffffffffffff. This can happen for example if `--data` specifies a directory.


tatsuhiro-t:

PR #182, opened 2019-10-28
Move template.h into server.cc

danbev:


tatsuhiro-t:

PR #183, opened 2019-10-29
Use ngtcp2_conn_state as type for state field

danbev:
This commit suggests updating `ngtcp2_conn`'s state field to be of type
`ngtcp2_conn_state` instead of type `int`.


tatsuhiro-t:

PR #185, opened 2019-10-31
Replace goto in create_ssl_ctx with exit

danbev:
This commit updates the server's `create_ssl_ctx` use `exit(EXIT_FAILURE)`
instead of using a goto label.

The motivation of this is to be consistent with `create_ssl_ctx` in
client.cc.


tatsuhiro-t:
Thank you.  Merged now.

PR #187, opened 2019-11-01
Introduce put_uint16_varint function

danbev:
This commit adds a static function named put_uint16_varint to reduce
some code duplication when putting some transport parameters.

I was not sure what would be preferred, returning `uint8_t*` like it's currently or making the function void. I'd be happy to change just let me know.

### DIFF
--- a/lib/ngtcp2_crypto.c
+++ b/lib/ngtcp2_crypto.c
@@ -88,7 +88,7 @@ static size_t varint_paramlen(uint64_t param) {
 
 static uint8_t *put_uint16_varint(uint8_t *p, ngtcp2_transport_param_id id,
                                   uint64_t value) {
-  p = ngtcp2_put_uint16be(p, id);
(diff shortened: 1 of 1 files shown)


lib/ngtcp2_crypto.c
```diff
@@ -86,6 +86,13 @@ static size_t varint_paramlen(uint64_t param) {
   return 4 + ngtcp2_put_varint_len(param);
 }
 
+static uint8_t *put_uint16_varint(uint8_t *p, ngtcp2_transport_param_id id,
+                                  uint64_t value) {
+  p = ngtcp2_put_uint16be(p, id);
```
tatsuhiro-t: It looks like we need to cast id to (uint16_t).

danbev: Sorry about that, I ran out of time on Friday. I'll update shortly, thanks!

tatsuhiro-t:
Thank you.  Merged now.

PR #195, opened 2019-11-05
Fix condition checking for Initial packet

amirlivneh:
The code previously determined whether a packet was an Initial packet by looking at its type value, without first checking it was not a Version Negotiation packet. In the case of a Version Negotiation packet, reading the type value returned the random value in the Unused field. This occasionally caused Version Negotiation packets to be misinterpreted as Initial packets. When this happened, ngtcp2_conn_read_pkt() returned NGTCP2_ERR_DRAINING instead of NGTCP2_ERR_RECV_VERSION_NEGOTIATION.

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -4702,7 +4702,11 @@ static int conn_recv_handshake_cpkt(ngtcp2_conn *conn, const ngtcp2_path *path,
       if (nread == NGTCP2_ERR_DISCARD_PKT) {
         goto fin;
       }
-      if (nread != NGTCP2_ERR_CRYPTO && (pkt[0] & NGTCP2_HEADER_FORM_BIT) &&
+      if (nread != NGTCP2_ERR_CRYPTO &&
+          (pkt[0] & NGTCP2_HEADER_FORM_BIT) &&
+          /* Not a Version Negotiation packet */
+          pktlen > 4 &&
+          ngtcp2_get_uint32(&pkt[1]) > 0 &&
           ngtcp2_pkt_get_type_long(pkt[0]) == NGTCP2_PKT_INITIAL) {
         goto fin;
       }


lib/ngtcp2_conn.c
```diff
@@ -4702,7 +4702,11 @@ static int conn_recv_handshake_cpkt(ngtcp2_conn *conn, const ngtcp2_path *path,
       if (nread == NGTCP2_ERR_DISCARD_PKT) {
         goto fin;
       }
-      if (nread != NGTCP2_ERR_CRYPTO && (pkt[0] & NGTCP2_HEADER_FORM_BIT) &&
+      assert(pktlen > 1);
+      if (nread != NGTCP2_ERR_CRYPTO &&
+          (pkt[0] & NGTCP2_HEADER_FORM_BIT) &&
```
tatsuhiro-t, line 4706: I think it would be safer to check that pktlen > 4 holds here. pktlen might be 1.

amirlivneh: Good point. Fixed.

tatsuhiro-t: Isn't it better to add pktlen > 4 in if condition rather than assert?  assertion may fail and it crashes application.

amirlivneh: I originally chose to assert because I expect conn_recv_handshake_pkt() to return NGTCP2_ERR_DISCARD_PKT if pktlen <= 5 because of the check in ngtcp2_pkt_decode_hd_long(). Therefore I assumed that if we reach this point with pktlen <= 4, it's necessarily a bug.

I now moved the check to the condition as suggested.

tatsuhiro-t: Thank you for explanation.  I think the assumption is ok for the current code.  Sorry I haven't checked the function.

PR #201, opened 2019-11-29
Fix use of uninitialized data in tests

amirlivneh:
write_single_frame_pkt() returns a packet length greater than the number of bytes it writes to the buffer. The returned length reflects a fake AEAD overhead, while this extra overhead is never written. As a result, ngtcp2_verify_stateless_reset_token() uses initialized data when reading the end of the packet. This causes noise when using a memory sanitizer.


tatsuhiro-t:

PR #202, opened 2019-12-09
Avoid use-after-free if key update is initiated while ppe is pending

amirlivneh:
conn_rotate_keys() frees conn->pktns->crypto.tx.ckm. If there is a pending ppe, the ppe object still holds a reference to the freed memory and will eventually try to use it in ngtcp2_ppe_final().


tatsuhiro-t:
Thank you.  Merged now.

PR #203, opened 2019-12-11
Fix example server's write callback never being called

amirlivneh:
Handler::wev_ is initialized with a file descriptor set to 0. ev_io_set() is never called for this watcher, so its file descriptor remains 0 and all calls to ev_io_start() have no effect. As a result writecb() is never called.

There may still be an issue due to all endpoints sharing the same watcher, but this is mostly orthogonal to the problem fixed here.

### DIFF
--- a/examples/server.cc
+++ b/examples/server.cc
@@ -1474,6 +1474,7 @@ int Handler::init(const Endpoint &ep, const sockaddr *sa, socklen_t salen,
     return -1;
   }
 
+  ev_io_set(&wev_, endpoint_->fd, EV_WRITE);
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Current code around wev_ is not so clean, but its only purpose at the moment is perform write in the next event loop, so any file descriptor is fine to associate it to wev_.  It might be easier to call `ev_io_set(&wev_, endpoint_->fd, EV_WRITE);` in Handler::init() for now.

amirlivneh:
Good idea. I updated the PR.

tatsuhiro-t:

PR #207, opened 2020-01-05
Additional crypto helper functions

jasnell:
Ended up implementing these entirely with ngtcp2_crypto, thought
they may make useful utility/helper additions:

* `ngtcp2_crypto_derive_retry_token_key` - Used to derive a key used
  in the generation of an encrypto retry token.

* `ngtcp2_crypto_generate_stateless_reset_token` - Used to generate
  a stateless reset token as an HKDF extraction using the CID and
  a token secret as input.


tatsuhiro-t:
Thank you for PR.  I decided to include ngtcp2_crypto_generate_stateless_reset_token only because I'm not sure that the way of generating token key in examples server is robust and suitable for general use at the moment.
I tweaked ngtcp2_crypto_generate_stateless_reset_token so that caller can specify message digest function.  See bdd640e7e63d5ea3ff0af9a63261a252949b71c5

jasnell:
Thank you :-)

PR #210, opened 2020-01-22
Make ngtcp2_pkt_write_stateless_reset use const for input

addaleax:
These two input buffers are not modified as part of the function
and can/should thus be `const`.


tatsuhiro-t:
Thank you.  Merged now.

PR #215, opened 2020-02-11
Remove redundant Buffer::begin

amirlivneh:
It always equals to buf.data().


tatsuhiro-t:

PR #217, opened 2020-02-26
Update AES-based header protection algorithm to use AES-ECB

ueno:
Until draft-16, the header protection mask was calculated using AES-CTR, while it has changed to use AES-ECB in draft-17. Although I suspect this is kept for interoperability reasons, it would be nice to use ECB as some crypto libs don't support CTR (single block ECB can be emulated with CBC, which is available everywhere).

### DIFF
--- a/crypto/openssl/openssl.c
+++ b/crypto/openssl/openssl.c
@@ -27,6 +27,7 @@
 #endif /* HAVE_CONFIG_H */
 
 #include <assert.h>
+#include <string.h>
 
 #include <ngtcp2/ngtcp2_crypto.h>
 #include <ngtcp2/ngtcp2_crypto_openssl.h>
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  It looks like you just changed implementation detail of the particular crypto library (openssl) and it supports AES-CTR.  If "some crypto libs" mean libraries other than openssl, then I think we just say in documentation that AES-ECB can be used here if AES-CTR is not available.

ueno:
Sorry, I should have added a bit more context. Let me rephrase: I'm trying to write a backend using GnuTLS, and realized that our AES-ECB based implementation doesn't interoperate with the stock `examples/server` compiled with the OpenSSL backend, because AES-CTR and AES-ECB are completely different algorithms. I'm not sure if it's sufficient to just document the difference.

ueno:
Sorry, it turned out to be completely my misunderstanding: indeed both calculate the same value, and the bug was in my backend code. I apologize for the noise.

PR #219, opened 2020-03-09
Not sending ack frames that acknowledge no blocks.

oesh:
Current implementation of ngtcp2_conn will send ack frames even when there are no ranges to acknowledge. This causes ngtcp2 to send empty ack frames. 

This diff treats an ACK frame with an empty ACK range as "no packets to acknowledge": 
- *pfr is untouched.
- the function returns zero.


oesh:
Need more work.

PR #222, opened 2020-03-18
Remove handling of NETWORK_ERR_SEND_BLOCKED in example server

amirlivneh:
The example server uses sendmsg() in blocking mode and NETWORK_ERR_SEND_BLOCKED is never returned as an error.


tatsuhiro-t:

PR #225, opened 2020-03-25
crypto: add GnuTLS backend

### LINKED ISSUE #216: GnuTLS crypto back-end
I would like to know if ngtcp2 maintainers would be interested in having a GnuTLS based crypto back-end.

Currently, changes are being made in GnuTLS to add the API's required to support QUIC:

* Exporting secrets from TLS: https://gitlab.com/gnutls/gnutls/-/merge_requests/1184
* Exposing HKDF API: https://gitlab.com/gnutls/gnutls/-/merge_requests/1186
* Add API's to push/get unprotected data from TLS: https://gitlab.com/gnutls/gnutls/-/merge_requests/1187

If there is interest, I can provide the required patches, although slightly intrusive changes would be required in ngtcp2's code as it is currently strongly tied to OpenSSL QUIC implementation from the modified branch (the examples use OpenSSL directly and the crypto helper library interface design assumes the TLS implementation received intrusive changes to be "QUIC aware", e.g. handling QUIC encryption levels internally).

ueno:
This adds a crypto backend based on GnuTLS.  While most of the `gnutls_*` functions used in this backend are officially available in upstream GnuTLS, the following functions are only available in the
['tmp-quic' branch](https://gitlab.com/gnutls/gnutls/-/commits/tmp-quic), for ABI assurance reasons until the QUIC standard is finalized:
- gnutls_handshake_write
- gnutls_quic_get_peer_transport_params

There is also a port of `ngtcp2/examples` in a separate repository:
https://github.com/ueno/ngtcp2-gnutls-examples

Currently, the `server` hits assertion failure inside `libngtcp2`, where it assumes that read/write secrets are provided at the same time, while GnuTLS internally does it asymmetrically for some reason (I guess the [new API in BoringSSL](https://github.com/google/boringssl/commit/1e859054c31ec7b974e998373e85fadea56d93ad) will also expose a similar problem).

Fixes #216.


tatsuhiro-t:
Thank you for PR.  I'll review this PR.
As for new API in BoringSSL, I expect that OpenSSL will follow the changes.  Before that, I'll change the ngtcp2 API to install 1RTT rx and tx keys separately so that ngtcp2 can work with both APIs including GNUTLS.

tatsuhiro-t:
Merged now.  Again, thank you for PR.

tatsuhiro-t:
32e70316461321df7b51e98450290f3818368b65 splits key installation function into 2 for each rx and tx.  I hope that it can fix assertion failure in GnuTLS enabled server.

tatsuhiro-t:
I found that it might be better to bump required GnuTLS version to 3.6.13 or later because 3.6.12 is released version and does not have new function used in this patch.

tatsuhiro-t:
FYI I made a commit 277206918e5839990b2467322b8c42a7a3a42e93 to fix compile error with clang-10 with the assumption that non-null **and** non-unknown cipher/hash is good to go.

ueno:
Thank you for merging, and updating the API!

> 32e7031 splits key installation function into 2 for each rx and tx. I hope that it can fix assertion failure in GnuTLS enabled server.

Thank you. After updating the gnutls branch and ngtcp2-gnutls-example, the server started working without assertion failure.

> I found that it might be better to bump required GnuTLS version to 3.6.13 or later because 3.6.12 is released version and does not have new function used in this patch.

Unfortunately, we currently don't use the "post-release version bump" scheme in GnuTLS, so even in the git master, the version is still 3.6.12. We have a rough plan to branch 3.7 after 3.6.13 (planned for this week), so it may possible to use that version as an indicator for QUIC support.

PR #230, opened 2020-04-20
Simplify remove_tx_stream_data()

amirlivneh:
The return value from this function is not used.


tatsuhiro-t:

PR #231, opened 2020-04-25
Fix format specifier in log_fr_connection_close()

amirlivneh:
ngtcp2_connection_close.reason_len is size_t, which is not necessarily 64-bit.


tatsuhiro-t:

PR #233, opened 2020-04-28
Remove unused string.h include from examplestest

danbev:


tatsuhiro-t:
Thank you.  Merged now.

PR #235, opened 2020-04-29
Remove unused errno.h include from ngtcp2_log

danbev:
Commit 0999d327a1511acdc3fed93d2c6d040b5256dbad ("Pass user-defined printf function rather than passing file descriptor") removed the usage of errno so it does not need to be included.


tatsuhiro-t:
Thank you.  Merged now.

PR #236, opened 2020-04-30
Fix integer wraparound on 32-bit machines

amirlivneh:
This causes conn_get_earliest_pktns() to falsely report a nonzero loss time, resulting in spurious retransmissions.


tatsuhiro-t:
Nice catch.  Thank you for PR.

Issue #238, opened 2020-05-04
The default build fails due to GnuTLS


Lekensteyn:
Following the instructions in the README will end up in a build failure because GnuTLS apparently needs special patches (see PR #225 by @ueno, issue #216).

Some suggestions to improve this situation:
 - [ ] Update README with instructions to build an appropriate GnuTLS backend.
 - [x] Build the OpenSSL backend only when a good GnuTLS version is missing. This would allow 3109444df8e83d7746f825dbf9fba61a168b82cd to be reverted.
 - [x] Update CI config to ensure that the GnuTLS has sufficient test coverage.
 - [x] Update the build systems to perform feature detection, so it fails early when OpenSSL or GnuTLS do not have suitable APIs.

I am using the CMake build system on Arch Linux which ships with GnuTLS 3.6.13.

Build output:
```
[28/51] Building C object crypto/gnutls/CMakeFiles/ngtcp2_crypto_gnutls.dir/gnutls.c.o
FAILED: crypto/gnutls/CMakeFiles/ngtcp2_crypto_gnutls.dir/gnutls.c.o 
...
crypto/gnutls/gnutls.c:260:8: error: unknown type name ‘gnutls_record_encryption_level_t’
crypto/gnutls/gnutls.c:264:12: error: ‘GNUTLS_ENCRYPTION_LEVEL_INITIAL’ undeclared (first use in this function); did you mean ‘NGTCP2_CRYPTO_LEVEL_INITIAL’?
crypto/gnutls/gnutls.c:264:12: note: each undeclared identifier is reported only once for each function it appears in
crypto/gnutls/gnutls.c:266:12: error: ‘GNUTLS_ENCRYPTION_LEVEL_HANDSHAKE’ undeclared (first use in this function); did you mean ‘NGTCP2_CRYPTO_LEVEL_HANDSHAKE’?
crypto/gnutls/gnutls.c:268:12: error: ‘GNUTLS_ENCRYPTION_LEVEL_APPLICATION’ undeclared (first use in this function)
crypto/gnutls/gnutls.c:270:12: error: ‘GNUTLS_ENCRYPTION_LEVEL_EARLY’ undeclared (first use in this function); did you mean ‘NGTCP2_CRYPTO_LEVEL_EARLY’?
crypto/gnutls/gnutls.c:283:9: warning: implicit declaration of function ‘gnutls_handshake_write’; did you mean ‘gnutls_handshake’? [-Wimplicit-function-declaration]
[...]

PR #239, opened 2020-05-04
Fix detection of proper OpenSSL and GnuTLS libraries

Lekensteyn:
Until mainline support of QUIC are present in OpenSSL and GnuTLS, be
sure to test whether the found library versions are the patched ones
since the version number check is insufficient. This fixes the default
build on a system where a new GnuTLS is present without the patches from
the GnuTLS tmp-quic branch.

Add new ENABLE_OPENSSL and ENABLE_GNUTLS CMake options to allow
explicitly disabling these libraries to mirror the --without-openssl and
the --without-gnutls functionality of autotools.

Partially addresses https://github.com/ngtcp2/ngtcp2/issues/238
___
Tested with CMake and autotools on Arch Linux with the patched OpenSSL version, it is picked up and examples are built. GnuTLS is correctly marked as not found and the build now passes.

I have not tried to test this with a patched GnuTLS since its build is quite complicated. @ueno could you give this a try?

To test this, I have set `PKG_CONFIG_PATH=/tmp/openssl/prefix/lib/pkgconfig:/tmp/nghttp3/prefix/lib/pkgconfig` prior to running `cmake` or `configure`. Adjust paths accordingly.


tatsuhiro-t:
Looks good.  Thank you for PR.

PR #240, opened 2020-05-08
Fix static library builds, clarify shared libs situation

Lekensteyn:
A functional fix and a documentation fix:

* CMake: ensure libs are built position-independent

  Static libraries have to be built as position-independent, otherwise
  they cannot be linked into a shared library which is built
  position-independent (-fPIC).

  Without this, linking objects into libcurl.so fails with:

      /usr/bin/ld: libngtcp2.a(ngtcp2_conn.c.o): relocation R_X86_64_PC32 against symbol `__asan_option_detect_stack_use_after_return' can not be used when making a shared object; recompile with -fPIC

* CMake: add BUILD_SHARED_LIBS option

  Prior to commit 4e955731d3a6ac344a593ee9d9c5d00982846adf, ngtcp2 was
  built as shared library. That has changed into a static or shared
  library depending on the BUILD_SHARED_LIBS option, but this option was
  not documented and the comments were misleading. Rectify both issues.

  Note that autotools builds both libraries by default, but the current
  CMake config is only able to build one of them.

__
cc @DaanDeMeyer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,6 +313,8 @@ message(STATUS "summary of build options:
       CXXFLAGS:       ${CMAKE_CXX_FLAGS_${_build_type}} ${CMAKE_CXX_FLAGS}
       WARNCFLAGS:     ${WARNCFLAGS}
       WARNCXXFLAGS:   ${WARNCXXFLAGS}
+    Library:
+      Shared:         ${BUILD_SHARED_LIBS}
     Test:
       CUnit:          ${HAVE_CUNIT} (LIBS='${CUNIT_LIBRARIES}')
     Libs:
--- a/CMakeOptions.txt
+++ b/CMakeOptions.txt
@@ -1,5 +1,7 @@
 # Features that can be enabled for cmake (see CMakeLists.txt)
 
+option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
+
 option(ENABLE_WERROR    "Make compiler warnings fatal" OFF)
 option(ENABLE_DEBUG     "Turn on debug output" OFF)
 option(ENABLE_ASAN      "Enable AddressSanitizer (ASAN)" OFF)
--- a/crypto/gnutls/CMakeLists.txt
+++ b/crypto/gnutls/CMakeLists.txt
@@ -47,6 +47,7 @@ set_target_properties(ngtcp2_crypto_gnutls PROPERTIES
   COMPILE_FLAGS "${WARNCFLAGS}"
   VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
   C_VISIBILITY_PRESET hidden
+  POSITION_INDEPENDENT_CODE ON
 )
 
 target_include_directories(ngtcp2_crypto_gnutls PUBLIC
--- a/crypto/gnutls/CMakeLists.txt
+++ b/crypto/gnutls/CMakeLists.txt
@@ -41,7 +41,7 @@ foreach(name libngtcp2_crypto_gnutls.pc)
   configure_file("${name}.in" "${name}" @ONLY)
 endforeach()
 
-# Public shared library
+# Public shared or static library
 add_library(ngtcp2_crypto_gnutls ${ngtcp2_crypto_gnutls_SOURCES})
 set_target_properties(ngtcp2_crypto_gnutls PROPERTIES
   COMPILE_FLAGS "${WARNCFLAGS}"
--- a/crypto/openssl/CMakeLists.txt
+++ b/crypto/openssl/CMakeLists.txt
@@ -41,7 +41,7 @@ foreach(name libngtcp2_crypto_openssl.pc)
   configure_file("${name}.in" "${name}" @ONLY)
 endforeach()
 
-# Public shared library
+# Public shared or static library
 add_library(ngtcp2_crypto_openssl ${ngtcp2_crypto_openssl_SOURCES})
 set_target_properties(ngtcp2_crypto_openssl PROPERTIES
   COMPILE_FLAGS "${WARNCFLAGS}"
--- a/crypto/openssl/CMakeLists.txt
+++ b/crypto/openssl/CMakeLists.txt
@@ -47,6 +47,7 @@ set_target_properties(ngtcp2_crypto_openssl PROPERTIES
   COMPILE_FLAGS "${WARNCFLAGS}"
   VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
   C_VISIBILITY_PRESET hidden
+  POSITION_INDEPENDENT_CODE ON
 )
 
 target_include_directories(ngtcp2_crypto_openssl PUBLIC
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -70,6 +70,7 @@ set_target_properties(ngtcp2 PROPERTIES
   COMPILE_FLAGS "${WARNCFLAGS}"
   VERSION ${LT_VERSION} SOVERSION ${LT_SOVERSION}
   C_VISIBILITY_PRESET hidden
+  POSITION_INDEPENDENT_CODE ON
 )
 
 target_include_directories(ngtcp2 PUBLIC ${ngtcp2_INCLUDE_DIRS})
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -64,7 +64,7 @@ set(ngtcp2_INCLUDE_DIRS
   "${CMAKE_CURRENT_BINARY_DIR}/includes"
 )
 
-# Public shared library
+# Public shared or static library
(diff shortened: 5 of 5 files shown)


Lekensteyn:
Interesting build errors on the cmake+macOS config, investigating:
```
[ 82%] Linking C shared library libngtcp2_crypto_openssl.dylib
Undefined symbols for architecture x86_64:
  "_EVP_CIPHER_CTX_ctrl", referenced from:
      _ngtcp2_crypto_encrypt in openssl.c.o
      _ngtcp2_crypto_decrypt in openssl.c.o
  "_EVP_CIPHER_CTX_free", referenced from:
      _ngtcp2_crypto_encrypt in openssl.c.o
      _ngtcp2_crypto_decrypt in openssl.c.o
      _ngtcp2_crypto_hp_mask in openssl.c.o
...
  "_ngtcp2_pkt_write_retry", referenced from:
      _ngtcp2_crypto_write_retry in shared.c.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [crypto/openssl/libngtcp2_crypto_openssl.0.0.0.dylib] Error 1
```

CMakeOptions.txt
```diff
@@ -1,5 +1,7 @@
 # Features that can be enabled for cmake (see CMakeLists.txt)
 
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
```
daandemeyer: `BUILD_SHARED_LIBS` is a CMake cache variable. I would not change the default value in the CMakeLists.txt itself, just pass `-DBUILD_SHARED_LIBS=ON` when invoking cmake.

CMakeLists.txt
```diff
@@ -258,6 +258,10 @@ if(ENABLE_DEBUG)
   set(DEBUGBUILD 1)
 endif()
 
+# Automatically enabled for shared libraries,
+# required for static library builds with ASAN.
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
daandemeyer: I'm not 100% sure but I would try setting this via a command line argument and see if that works before hardcoding it in the CMake file. `-DCMAKE_POSITION_INDEPENT_CODE=ON`

CMakeOptions.txt
Lekensteyn: It is not a default cache variable, it is supposed to be set by projects. And by making it an `option` as opposed to `set`, users can see the option when they use `ccmake` or `cmake-gui`.

From https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS
> This variable is often added to projects as an option()

The line is from this tutorial: https://cmake.org/cmake/help/latest/guide/tutorial/index.html#mixing-static-and-shared-step-9

CMakeLists.txt
Lekensteyn: These variables are not cache variables such as `CMAKE_BUILD_TYPE` and if I am not mistaken they are supposed to be set by the developer as opposed to the user.

Setting this option has two effects, enabling -fPIC for all libraries, and enabling PIE for executables (often seen as a hardening feature). I think there is value in enabling it by default (it changes the behavior for static libs and executables), but clearly some more work is needed as the macOS build somehow fails.

Docs for this option: https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html

Lekensteyn:
Thanks for your feedback Daan!

> If curl uses add_subdirectory to build ngtcp2, it will automatically inherit all the options and variables from curl's CMakeLists.txt so you can just configure everything there.

curl does not include ngtcp2 via `add_library`, instead ngtcp2 is built and installed as a normal library:
```sh
PKG_CONFIG_PATH=/tmp/openssl/prefix/lib/pkgconfig:/tmp/nghttp3/prefix/lib/pkgconfig \
cmake -GNinja ~/projects/ngtcp2 \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_INSTALL_PREFIX=$PWD/prefix \
    -DCMAKE_INSTALL_LIBDIR=lib \
    -DENABE_ASAN=1
ninja
ninja install
```

CMakeOptions.txt
daandemeyer: You're right, I'm wrong, it isn't a cache variable. However, it doesn't behave any differently whether you specify it as an option or not (It always keeps its value over different invocations). You're also changing the default behaviour of CMake in that by default, libraries are built as static libraries. This is mostly preference but as a user of many CMake projects, I prefer them to stick to the defaults so they all behave the same (Some projects build both types by default which is even worse to deal with). Static libraries is the default so when building a new project I naturally expect to get static libraries.

I don't mind keeping the option, but I would default to `OFF` if only to stick to CMake's defaults.

CMakeLists.txt
daandemeyer: The docs for this stuff really need to be updated. I just tried it out on my own library, and specifying `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` via a command line argument (using static libraries) works and keeps its value over multiple CMake invocations so it effectively behaves like an option.

The reasoning is again the same, I'd prefer to not mess with defaults as much as possible. This makes it much easier to compile many CMake projects uniformly via passing `CMAKE_CXX_FLAGS` or similar stuff. In general, I think projects should by default compile with the most minimal set of compilation flags possible so you're not forcing anything on users. Each project setting the defaults for stuff like this themselves is just a recipe for disaster. Instead, do the least amount of stuff possible so global defaults can be applied for many projects by distros and packagers without annoying issues.

Of course, adding off by default options that enable all the good stuff is not a problem. For an example, see [reproc](https://github.com/DaanDeMeyer/reproc/blob/master/cmake/reproc.cmake) that has a bunch of options, most of them off by default that enable all sorts of good stuff. However, by default, it adds no compile flags at all in order to make it as easy as possible for distros and whatnot to configure their own set of flags.

This turned into a bit longer of a response than I expected. Anyway, all of this stuff is just my own views on the subject. Feel free to ignore if you disagree.

CMakeOptions.txt
Lekensteyn: It looks like we have two use cases with different expecations :)

Public libraries (installed in /usr/lib) are usually installed as shared libraries. Static libraries are not always available, especially on Debian/Ubuntu. The current autotools config in ngtcp2 does build both the shared and static libs.

Your use case presumably embeds ngtcp2 in another project which expects ngtcp2 to stay internal, hence static linking. I added a static library earlier for unittests (`ngtcp2_static`), but this could presumably be used by your project as well. It will probably need a special build option (e.g. `ENABLE_STATIC` or whatever is common for CMake projects).

I'll have a look at the common practices for CMake projects and try to come up with something familiar.

CMakeLists.txt
Lekensteyn: If you specify `-Dx=y` through the `cmake` invocation, it will be stored in the cache (as if [`set(x y CACHE STRING "..."`](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry) was called. The difference between a plain variable and an "option" is that "option" creates a cache entry by default such that you can select it through `ccmake` or `cmake-gui`.

I think I agree with your viewpoint. It is somewhat annoying if projects don't follow CMake standards and add `-O2 -DNDEBUG` even when using `-DCMAKE_BUILD_TYPE=None`. For top-level projects, setting `CMAKE_POSITION_INDEPENDENT_CODE` might seem reasonable, but if it is included via `add_subdirectory` as used in [bnl](https://github.com/DaanDeMeyer/bnl/blob/master/cmake/bnl.cmake#L290), it will probably indeed override stuff from the parent project. To handle such embedded cases better, I'l look into target-specific options.

daandemeyer: Maybe put this behind the `ENABLE_ASAN` option as well? I'm not sure if you need `-fPIC` when not compiling with sanitizers.

Lekensteyn:
Replaced the original change of defaults (static->shared) with a new functional fix and another documentation fix.

CMakeLists.txt
Lekensteyn: Have a look at the new commit message, I believe that `-fPIC` is necessary whenever you try to combine a static library (ngtcp2) into another shared library (which is always built with `-fPIC`).

DaanDeMeyer:
LGTM, thanks for taking my feedback into account. (you don't necessarily have to tag me for ngtcp2 related CMake stuff, I haven't been involved for quite a while. I'm always happy to provide feedback though, should you want it)

Lekensteyn:
Thanks! Usually I tag the person who made a related change (in this case, the change from always-shared -> static-by-default). Your feedback has been very helpful in understanding why the original change was made, and why my original "fix" could be improved. So if I run into a similar behavioral change, I hope I can poke you again :-)

tatsuhiro-t:
Thank you all for this PR, merged now.

PR #241, opened 2020-05-11
Fix read-heap-buffer-overflow in decode_varint_param()

amirlivneh:
If `valuelen` is 0 and `p` equals `end`, `decode_varint_param()` reads an extra byte past the end of the buffer.


tatsuhiro-t:

PR #242, opened 2020-05-24
gnutls: add stub ngtcp2_crypto_set_local_transport_params

ueno:
Reported by Peter Wu in:
https://github.com/curl/curl/pull/5443#issuecomment-633138665


ueno:
I believe this fixes the build, though I haven't managed to make the example server work with the new API (the client works). Let me look into it this week.

tatsuhiro-t:

PR #243, opened 2020-05-24
Fix potential wraparound in conn_update_recv_rate()

amirlivneh:
This is unlikely to wraparound in the real world, but fixing it makes the code easier to fuzz.


tatsuhiro-t:
Nice catch.  Thank you.  Merged now.

PR #244, opened 2020-06-17
Fully initialize ngtcp2_transport_params

addaleax:
Node.js makes the full `ngtcp2_transport_params` struct available
for storage as a byte array, including any possible padding
in the struct.

Fully initializing the struct with `memset()` makes this deterministic.


tatsuhiro-t:

PR #247, opened 2020-06-23
The gcc builtin __builtin_clzll is not available in Windows

jasnell:
The gcc builtin __builtin_clzll is not available in Windows when building using windows.

### DIFF
--- a/lib/ngtcp2_cc.c
+++ b/lib/ngtcp2_cc.c
@@ -31,6 +31,15 @@
 #include "ngtcp2_mem.h"
 #include "ngtcp2_rcvry.h"
 
+#ifdef _MSC_VER
+// __builtin_clzll is not available by default in _MSC_VER but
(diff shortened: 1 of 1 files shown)


jasnell:
Even this really isn't ideal for our case since we need to be able to build and compile on x86. Current solution on x86 is to define `__builtin_clzll()` as a non-op function and we'll block use of the `cubic` congestion method on that platform unless there's a replacement here that would allow it to compile and function correctly on win x86

tatsuhiro-t:
@jasnell Thank you for PR.  #248 has workaround for windows x86 (32 bits build) which lacks __lzcnt64.  Does it work for you?

jasnell:
Yes, #248 should work just fine. Thank you :)

tatsuhiro-t:
Thank you.  I'll merge #248 and close this PR.

PR #248, opened 2020-06-24
Fix MSVC build

jblazquez:
`ngtcp2_cc.c` is using the `__builtin_clzll` intrinsic which doesn't exist on MSVC. Added the use of the corresponding [`__lzcnt` intrinsics](https://docs.microsoft.com/en-us/cpp/intrinsics/lzcnt16-lzcnt-lzcnt64?view=vs-2019) for MSVC.


tatsuhiro-t:

Issue #254, opened 2020-08-12
sys/socket.h not available on windows


jasnell:
https://github.com/ngtcp2/ngtcp2/commit/f18bb8544137d9d796902e08b0942d814f8a7474 breaks building on windows given that `sys/socket.h` (used in `ngtcp2.h`) and `netinet/ip.h` (used in `lib/ngtcp2_addr.c`) are not available (we need `winsock2.h` instead). I'll try to open a PR soon.

PR #255, opened 2020-08-12
Use winsock.h on windows instead of sys/socket.h

jasnell:
Fixes: #254


jasnell:
So the changes are also failing on freebsd. Please hold off landing this until I can update it to cover that platform as well :-)

jasnell:
Ok, I've updated the PR with include headers that appear to work on all platforms we support. Specifically, without these the build was failing on Windows and FreeBSD.

tatsuhiro-t:
Thank you for PR!  Merged now.

PR #273, opened 2020-10-12
Fix inconsistent "defined()" operator usage which causes compile warn…

lnslbrty:
…ings if undefined.

Signed-off-by: Toni Uhlig <matzeton@googlemail.com>


tatsuhiro-t:

PR #315, opened 2021-08-29
doc: Clarify NGTCP2_WRITE_STREAM_FLAG_MORE behavior on how to stop coalescing

ueno:
I was trying to write a small program using `NGTCP2_WRITE_STREAM_FLAG_MORE`, and stumbled on how to stop the loop if there is no data to be fed. Hopefully this makes it a little clearer.

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -3458,7 +3458,7 @@ NGTCP2_EXTERN int ngtcp2_conn_read_pkt(ngtcp2_conn *conn,
  * @function
  *
  * `ngtcp2_conn_write_pkt` is equivalent to calling
- * `ngtcp2_conn_writev_stream` without specifying stream data and
+ * `ngtcp2_conn_writev_stream` with -1 as stream_id, no stream data, and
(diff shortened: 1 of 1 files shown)


lib/includes/ngtcp2/ngtcp2.h
```diff
@@ -4091,7 +4091,8 @@ NGTCP2_EXTERN ngtcp2_ssize ngtcp2_conn_write_stream(
  * function).  Just keep calling `ngtcp2_conn_writev_stream`,
  * `ngtcp2_conn_write_pkt`, or `ngtcp2_conn_writev_datagram` until it
  * returns a positive number (which indicates a complete packet is
- * ready).
+ * ready).  If there is no stream data to include, call
+ * `ngtcp2_conn_write_pkt` to stop coalescing and write a packet.
```
tatsuhiro-t: I think it would be nice to first recommend ngtcp2_conn_writev_stream with  stream_id = -1 and no stream data, as you described in the above hunk because it is more natural to use same function in a loop.

ueno: Thank you; updated along these lines.

tatsuhiro-t:

PR #322, opened 2021-09-07
we do not care about result of this path validation

Natris:
The NGTCP2_PV_FLAG_DONT_CARE path validation is there to return to the old path in case of fradulent migration (it would be cleaner if this would run in parallel with the path validation triggered by migration, but that would require parallel path validations, cough cough cough :) ); we should not report it's result via callback


tatsuhiro-t:

PR #326, opened 2021-09-17
Fix incorrect return value from ngtcp2_pkt_decode_datagram_frame()

amirlivneh:
The value returned from ngtcp2_pkt_decode_datagram_frame() is later returned from ngtcp2_pkt_decode_frame(). According to its contract, ngtcp2_pkt_decode_frame() returns NGTCP2_ERR_FRAME_ENCODING when the payload is too short.


tatsuhiro-t:

PR #329, opened 2021-10-05
Suppress compiler warnings for NDEBUG

danbev:
Currently when configuring using the following options there are a
number of compiler warnings generated:
```console
$ env CFLAGS='-DNDEBUG -O3' ./configure --enable-debug=no --disable-silent-rules --enable-werror PKG_CONFIG_PATH=$PWD/../openssl/build/lib/pkgconfig:$PWD/../nghttp3/build/lib/pkgconfig LDFLAGS="-Wl,-rpath,$PWD/../openssl/build/lib"
```
This commit contains suggestions to suppress these warnings.

### DIFF
--- a/crypto/openssl/openssl.c
+++ b/crypto/openssl/openssl.c
@@ -53,6 +53,7 @@ static size_t crypto_aead_max_overhead(const EVP_CIPHER *aead) {
     return EVP_CCM_TLS_TAG_LEN;
   default:
     assert(0);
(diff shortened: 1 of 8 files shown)


tatsuhiro-t:
Thank you for PR.  Cherry-picked as a9f72c49854deb1387c0dcbac7c0cd5fd3f76c34 with the following changes:
- abort() consistently instead of initializing local variables
- place (void)foo after variable definitions

PR #331, opened 2021-10-12
Updated the version of `quictls/openssl`

D4v1dH03:
Additional, some situations, `make install` need to be prepended with `sudo` otherwise it will result in a `Permission denied` error

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,7 +65,7 @@ Build from git
 
 .. code-block:: text
 
-   $ git clone --depth 1 -b openssl-3.0.0+quic https://github.com/quictls/openssl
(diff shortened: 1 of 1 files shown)


D4v1dH03:
See related pull request on curl/curl#7842.

D4v1dH03:
Can @tatsuhiro-t please look at this?

tatsuhiro-t:
Thank you for PR.  OpenSSL versions were updated via 20c710a8789ec910455ae4e588c72e9e39f8cec9
I'd like to keep openssl 1.1.1 default because openssl 3.0.1 still has a performance issue.

PR #343, opened 2021-12-31
Check ssl is not null

flx413:
SSL_CTX_new may be return NULL in low memory case.


tatsuhiro-t:

Issue #337, opened 2021-11-09
Alignment specifier does not work as expected


jpboivin:
Hello,

This commit (https://github.com/ngtcp2/ngtcp2/commit/00713cb932f7d55f17aa6bfdfbc2de813e7b5494) introduced alignment specifiers on some structures. It seems that it does not work as expected when combined with the `auto` keyword in C++ on GCC and Clang. It would silently compile with earlier versions of Clang, but Clang 13 introduced a new warning so at least it is detected at compile-time now (`-Walign-mismatch`) instead of being UB.

If you write the following code, the alignment of the variable will be the natural alignment (4) even if the alignment of the type seems to be 8.
```
const auto packetInfo2 = ngtcp2_pkt_info{};
```

Digging a bit around, I think the issue is that the alignment is applied to the type alias, and not the underlying structure. `auto` probably uses the underlying structure when doing type deduction. In C++, `alignas` is prohibited in type aliases (both `using` and `typedef`), so it seems to be a corner case that isn't properly supported by the `__attribute__` even if it compiles fine. I made a small code snippet that reproduce the issue: https://godbolt.org/z/19h41qncq

The fix would be to add the alignment specifier before, like for MSVC. With that said, I'm no C expert, so I'm not sure if is right or wrong in that language. I'd tend to say it should be right too, but I might be missing something.

Note. Behavior is the same with GCC or earlier Clang versions. `auto` variables will not respect the alignment specifier if it is after the type. It just happens that misaligned pointer params are not detected at compile-time with a warning.

tatsuhiro-t:
Thank you for letting me know.  The behaviour around auto is quite interesting.  I missed the text in gcc documentation that attribute can be put after struct keyword.  All the examples put it after the '}'.  Will fix it.

tatsuhiro-t:
Fix committed via cc97643c3ce32e5f2c84270fd878ab324cde9600

AhmedShah63:
cc97643

PR #348, opened 2022-01-20
Fix the non-standard alignment specifier for C++

jpboivin:
This is an additional fix for https://github.com/ngtcp2/ngtcp2/issues/337, but this one for MSVC.

----

Microsoft introduced the C5029 warning with Visual Studio 2015. This warning only applies to C++ mode, but it does not seem to be language specific (i.e. ngtcp2 will trigger it when included from C++, even if it is within an extern C block). Basically, C5029 will be raised when an alignment specified is added to an alias and not the original type, because this is a non-standard extension (see [this page](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4800-through-c4999?view=msvc-170)).

```
typedef __declspec(align(8)) struct ngtcp2_pkt_info {
  // ...
} ngtcp2_pkt_info;
```

In the code you had previously, the `ngtcp2_pkt_info` structure was aligned with natural alignment, while the `ngtcp2_pkt_info` alias was overaligned to 8 bytes. This behavior is documented [here](https://docs.microsoft.com/en-us/cpp/cpp/align-cpp?view=msvc-170#vclrf_declspecaligntypedef).

As clearly there is no intent to have the alias and the type with different alignments (as you've already fixed for GCC and Clang), I did the same change for MSVC.

----

This MR also disable the [C4324](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4324?view=msvc-170) warning for Windows platforms. As you are overaligning the structure, padding will be added. It is fine by itself, but as it is part of the public header, it is best if this warning is disabled to avoid the users do it.

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -32,6 +32,11 @@
 #  define WIN32
 #endif
 
+#ifdef WIN32
+#  pragma warning(push)
+#  pragma warning(disable : 4324)
+#endif
+
 #include <stdlib.h>
 #if defined(_MSC_VER) && (_MSC_VER < 1800)
 /* MSVC < 2013 does not have inttypes.h because it is not C99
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:

PR #350, opened 2022-01-21
Make static libs symbols hidden

jpboivin:
This commit (https://github.com/ngtcp2/ngtcp2/commit/5ad56ded717da5d01cc802816ddd1bbae1a99957) made a change to the main targets to make them no longer based on CMake's `BUILD_SHARED_LIBS` option. With that change, the proper target to use for a static library variant of ngtcp2 is the `ngtcp2_static` target, but its properties are slightly different than previously.

1. Symbols are no longer hidden. On some platforms (i.e. GNU/Linux), linking a static library with _exported_ symbols into an executable or a shared library will re-export these symbols. This mean you get an ELF or a SO with the ngtcp2 symbols exported. It is good practice I think to make the symbols hidden even for static libraries (and actually needed in some cases).
2. Library is no longer compiled with `-fPIC`, which might be needed in some cases when linked into an executable or a shared library (depends on the parent compilation flags, but little harm to always enable it).

I've explicitly re-added these options to all the static targets, so they'll behave as they did previously.


tatsuhiro-t:
libngtcp2_static lib still exports internal functions with this PR, at least on Linux.  Does  C_VISIBILITY_PRESET  also work for static library as well?  Not sure the situation on windows or other platforms.  The static lib is used to run unit test, so if it hides non public functions, unit tests should fail, but it works.

As for -fPIC, it might be better to default off because if we enable it statically, then one is unable to disable it.

jpboivin:
> libngtcp2_static lib still exports internal functions with this PR, at least on Linux. Does C_VISIBILITY_PRESET also work for static library as well? Not sure the situation on windows or other platforms. The static lib is used to run unit test, so if it hides non public functions, unit tests should fail, but it works.

Symbols will remain visible to the linker, as after all, it is a static library. What changes is the symbols visibility in the final executable or shared library. For example, with a default visibility, internal symbols of ngtcp2 will be exported by the executable or the shared library that links the static lib, even if they are not intended to be exported. I didn't find actual documentation like I did in the past when I faced a similar issue with OpenSSL, but you can see similar bug reports for OpenJPEG (https://github.com/uclouvain/openjpeg/issues/354).

As far as I know, this is a Unix/Linux-specific issue. Windows does not have this issue as symbols must be explicitly annotated to be exported (or you need to generate a `.def` file like Make does with `WINDOWS_EXPORT_ALL_SYMBOLS`).

> As for -fPIC, it might be better to default off because if we enable it statically, then one is unable to disable it.

What I wonder is, why would there be a need to disable `-fPIC`? It will be needed if the static library is linked in a shared library (rare but happens), or if it is linked in a position independent executable (which is starting to be more and more common nowadays). So, it seems to me that it would be a decent default, compared to compile the static library without `-fPIC`. There is no harm to link a position independent static library in other contexts, and as far as I know, the only drawback is that the codegen is slightly worse. 

In all cases, I guess it could be a CMake option if you want to keep it configurable?

tatsuhiro-t:
Thank you.  Now I understand the symbol visibility in static library.

-fPIC is disabled by default in build tools, and it is easy to enable it by supplying -DCMAKE_POSITION_INDEPENDENT_CODE=ON.  I'd like to leave it off and let user decide.

jpboivin:
> -fPIC is disabled by default in build tools, and it is easy to enable it by supplying -DCMAKE_POSITION_INDEPENDENT_CODE=ON. I'd like to leave it off and let user decide.

I reverted these changes

tatsuhiro-t:
Thank you.  Merged now.

Issue #276, opened 2020-10-19
Usage of struct sockaddr


jpboivin:
Hello, 

So far, ngtcp2 has been extremely easy to port to the ~10 platforms supported where I work because it doesn't rely on anything outside of plain C. Platform-specific details (like sockets, cryptography, file system) are normally kept out of the library. 

Since the following commit https://github.com/ngtcp2/ngtcp2/commit/f18bb8544137d9d796902e08b0942d814f8a7474, the path holds an `struct sockaddr` instead of the previous opaque structure. I can understand the better comparisons (I had to deal with padding issues with the custom structure we introduced), but I'm wondering if ngtcp2 shouldn't instead introduce its own structure for that? Although `struct sockaddr` is normally supported (and its IPv4/IPv6 equivalents), not all platforms have them.

tatsuhiro-t:
It might be nice to define struct sockaddr family for those platforms which lack these structs.
For platforms that have struct sockaddr, using them is extremely easy because they are provided by syscall and no conversion is necessary.

jpboivin:
I wouldn't mind defining them, but I wouldn't like defining them to generic `struct sockaddr` either. That's why maybe a ngtcp2_sockaddr would make sense (maybe a type alias on the platforms that support it). Currently, I would need to define these types (and the `AF_INET` defines) in the ngtcp2.h header, which I find hacky.

tatsuhiro-t:
ngtcp2_sockaddr (and sockaddr_* variants) might be a good idea.
I think enumerating platforms which have struct sockaddr (or platform not having them) is also tedious (we cannot use config.h here), it might be easier to add macro to indicate that it uses generic version sockaddr, like so:

```c
#ifdef USE_GENERIC_SOCKADDR
struct ngtcp2_sockaddr {
  ...
} ngtcp2_sockaddr;
...
#else /* !USE_GENERIC_SOCKADDR */
typedef struct sockaddr ngtcp2_sockaddr;
typedef struct sockaddr_storage ngtcp2_sockaddr_storage;
#endif
```

tatsuhiro-t:
Fixed by #354

PR #354, opened 2022-01-24
Add generic sockaddr types to support non-standard sockaddr platforms.

dagf2101:
This pull request address the following [issue](https://github.com/ngtcp2/ngtcp2/issues/276).
Some platforms have non-standard sockaddress types and ngtcp2 canot be ported as easily on these platforms.

To overcome this problem we have added new generic ngtcp2_sockaddr* structures that can be enabled using NGTCP2_USE_GENERIC_SOCKADDR or NGTCP2_USE_GENERIC_IPV6_SOCKADDR defines.

By default the new ngtcp2_sockaddr* are typedefs on the regular sockaddr* structures.

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -50,10 +50,26 @@
 #include <stdarg.h>
 #include <stddef.h>
 
-#ifdef WIN32
-#  include <winsock2.h>
+#ifndef NGTCP2_USE_GENERIC_SOCKADDR
+#  ifdef WIN32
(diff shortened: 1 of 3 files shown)


tatsuhiro-t:
Cherry-picked as 2b40eb7fa5cc19072317e6ceab6ec71d1bf526be
I made a following changes in the follow-up commits:
- Match ngtcp2_sockaddr_storage POSIX specification, setting 128 bytes long and align at 8 bytes boundary.
- Define ngtcp2_in_addr and ngtcp2_socklen to fix compile issues.

PR #372, opened 2022-02-18
Use ngtcp2_ssize instead of ssize_t

rhoxn:
There was an unintentional use of ssize_t, so I fixed it to use ngtcp2_ssize.


tatsuhiro-t:

PR #373, opened 2022-02-20
configure.ac: fix have_openssl_quic detection

jay:
Prior to this change OpenSSL >= 1.1.1 without QUIC was not properly
disabled. make would attempt to build libngtcp2_crypto_openssl anyway
and fail.

Bug: https://github.com/curl/curl/pull/8485
Reported-by: Michał Antoniak

Closes #xxxx


tatsuhiro-t:

PR #377, opened 2022-02-28
Fix regex replace of compiler flags

donny-dont:
In MSVC the regex replace of compiler flags for a release build caused some flags to be fused together, `/Gy/Z7`, which resulted in the following error.

`cl : Command line error D8016 : '/O2' and '/GZ' command-line options are incompatible`

Took the same section from nghttp2's CMakeLists which contains a space to fix the issue.


tatsuhiro-t:

PR #378, opened 2022-02-28
Set install locations when building shared libs

donny-dont:
Currently when building on Windows the dll and lib files are in the same location. This explicitly states where to install each part of the shared library.


donny-dont:
As a corollary @tatsuhiro-t I see that there aren't targets created when `ENABLE_SHARED_LIBS` is `ON` for the `boringssl` and `picotls` crypto backends. I ran into this when trying to hook in `ngtcp2` and `nghttp3` with `curl` using `boringssl`.

I have a patch locally for the `boringssl` one but wasn't sure if there's a reason why there wasn't a `SHARED` target for it.

tatsuhiro-t:
The reason behind it is that they both are static library and  are both version-less product (probably just for companies consumption only, google and fastly).
It seems that boringssl has an option to build shared library, but when I wrote the build file, it did not or I did not notice it.
So basically, you need to compile executable with the same static library.  The one of the advantage of shared library is its versioning, but if it depends on version-less static lib, then that advantage goes away.  I do not want to include bunch of boringssl or picotls symbols in ngtcp2 crypto helper libraries.  That means they have unresolved symbols and some platforms do not like that (i.e., windows).

tatsuhiro-t:
Now merged.  Thank you.

PR #424, opened 2022-04-19
Minor config changes to simpleclient

roticv:
I was playing around with using simpleclient to connect to meta servers and
found some incorrectly configured settings:
* The quic transport is mismatched between openssl layer and ngtcp2 causing meta servers to reject with fizz exception.
* The default max streams for bidi is 0 and meta server was rejecting with STOP_SENDING frame.

### DIFF
--- a/examples/simpleclient.c
+++ b/examples/simpleclient.c
@@ -250,6 +250,9 @@ static int client_ssl_init(struct client *c) {
     SSL_set_tlsext_host_name(c->ssl, REMOTE_HOST);
   }
 
+  /* For NGTCP2_PROTO_VER_V1 */
+  SSL_set_quic_transport_version(c->ssl, TLSEXT_TYPE_quic_transport_parameters);
+
   return 0;
 }
 
@@ -405,6 +408,7 @@ static int client_quic_init(struct client *c,
   ngtcp2_transport_params_default(&params);
 
   params.initial_max_streams_uni = 3;
+  params.initial_max_streams_bidi = 3;
   params.initial_max_stream_data_bidi_local = 128 * 1024;
   params.initial_max_data = 1024 * 1024;
 


tatsuhiro-t:
Setting transport parameter version explicitly is good one.  Changing the number of bidi streams is a little bit questionable.  What application protocol "meta" server is speaking?  And why did it send STOP_SENDING and to which stream?

roticv:
I was testing with the following:
* www.facebook.com, with alpn {0x02, 'h', '3'}
* live-upload.facebook.com, with alpn {0x4, 'r', 'u', 's', 'h'} based on https://www.ietf.org/id/draft-kpugin-rush-01.html

> Changing the number of bidi streams is a little bit questionable.

My understanding is that currently the default initial_max_streams_bidi is 0 (based on https://github.com/ngtcp2/ngtcp2/blob/main/lib/ngtcp2_conn.c#L13170) and this value is validated on mvfst server implementation (https://github.com/facebookincubator/mvfst/blob/main/quic/state/QuicStreamManager.cpp#L430).

In the simpleclient.c, ngtcp2_conn_open_bidi_stream(...) is used (instead of ngtcp2_conn_open_uni_stream) and so it makes sense to set initial_max_streams_bidi.

tatsuhiro-t:
Client initial_max_streams_bidi is the maximum number of bidi streams that server can open.  Client only cares about  initial_max_streams_bidi from server when opening bidi stream.

I think the reason why facebook server sent STOP_SENDING is that it did not like the message in the stream.  Simpleclient sends something like "GET /", but if server uses h3, the message needs to be h3 frames.

tatsuhiro-t:
Cherry picked partially in fd14d017a836eea945a13d49d01b656ddeb47915

PR #465, opened 2022-06-09
Disable the internal session cache if TLS session is stored externally

msoxzw:
Since TLS session has been stored externally, I feel that there is no need to enable session caching in OpenSSL and BoringSSL, just like [what Google QUICHE does](https://github.com/google/quiche/blob/74aa2867de8e650809d85cd6454aca4200ec6e36/quiche/quic/core/crypto/tls_client_connection.cc#L24-L27).


tatsuhiro-t:

PR #478, opened 2022-06-28
Define WIN32_LEAN_AND_MEAN before ws2tcpip.h

donny-dont:
Including `ws2tcpip.h` ends up including `wincrypt.h` which defines symbols which collide with OpenSSL structures, such as `X509_NAME`, unless `WIN32_LEAN_AND_MEAN` is present.

### DIFF
--- a/crypto/includes/ngtcp2/ngtcp2_crypto.h
+++ b/crypto/includes/ngtcp2/ngtcp2_crypto.h
@@ -32,6 +32,9 @@ extern "C" {
 #endif
 
 #ifdef WIN32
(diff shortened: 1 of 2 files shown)


donny-dont:
This was found when compiling the BoringSSL backend on Windows. It would fail to compile `boringssl.c` with errors pointing at this issue.

tatsuhiro-t:
I think it is reasonable.  If we have any problems, we can revisit this.

tatsuhiro-t:
Thank you for the PR.  Merged now.

PR #479, opened 2022-06-28
Add a shared ngtcp2_crypto_boringssl target

donny-dont:
Both the openssl and gnutls variants have a shared target but it wasn't present for boringssl.

### DIFF
--- a/crypto/boringssl/CMakeLists.txt
+++ b/crypto/boringssl/CMakeLists.txt
@@ -43,6 +43,26 @@ foreach(name libngtcp2_crypto_boringssl.pc)
   configure_file("${name}.in" "${name}" @ONLY)
 endforeach()
 
+# Public shared library
+if(ENABLE_SHARED_LIB)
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
boringssl shared lib is explicitly omitted because it does not have version.  Shared object without version is rather hard to use.

donny-dont:
This was done in the context of a Windows build which doesn't have the versioning issue. Not sure if that would influence your decision at all.

Related is https://github.com/curl/curl/pull/9065 which detects in the CMake of curl.

tatsuhiro-t:
I think static linking is better for non-versioned library because it cannot be safely reused, which means just linking statically is better approach.

tatsuhiro-t:
We have a similar discussion about boringssl shared lib before: https://github.com/ngtcp2/ngtcp2/pull/378

donny-dont:
Yea just putting it up here if there is interest. If not feel free to close it.

PR #480, opened 2022-06-29
fix ngtcp2_htons for windows

rhoxn:
Fixed a problem with endian conversion in windows.
There was a problem with the lower octet always being zero.


bagder:
This fixes the curl problem discussed here: https://github.com/curl/curl/issues/9053

tatsuhiro-t:
Thank you for PR.  Yes, this works for me too. Still tiny 0 bytes UDP datagram issue exists probably it is caused by something else.

PR #489, opened 2022-07-09
Fix bug that stream_close is not called when STOP_SENDING is cancelled

tatsuhiro-t:


tatsuhiro-t:
Needs #490 and additional ack in test.

PR #501, opened 2022-07-30
Always emit ACK immediately after receiving PING frame.

atlesn:
I discovered that the library does not always immediately emit an ACK frame when a PING frame is received. The default ACK threshold of 2 causes ACK to be emitted only for every other received PING frame when a connection is idling.

This patch will have the library always emit ACK whenever PING is received. I can see that setting the ACK threshold to 1 also works, but this has other implications.

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -8750,6 +8750,7 @@ static ngtcp2_ssize conn_recv_pkt(ngtcp2_conn *conn, const ngtcp2_path *path,
   ngtcp2_max_frame mfr;
   ngtcp2_frame *fr = &mfr.fr;
   int require_ack = 0;
+  int require_immediate_ack = 0;
   ngtcp2_crypto_aead *aead;
   ngtcp2_crypto_cipher *hp;
   ngtcp2_crypto_km *ckm;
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
If I remembered it correctly, RFC 9000 does not mandate immediate ACK for PING frame.

atlesn:
Yes that's true. It states that `The receiver of a PING frame simply needs to acknowledge the packet containing this frame.`, but not that the ACK must be sent immediately.

tatsuhiro-t:
So could you close this PR?  I do not see this improves anything.

atlesn:
I think it's useful to respond immediately to a PING frame since the whole point of them is to ensure a connection is kept alive. Is there some other solution to this, how is keep alive usually implemented with QUIC?

tatsuhiro-t:
PING just elicits acknowledgement.  That is enough to keep connection alive.  All we need is acknowledgement, not immediate response. Sending a packet to keep connection alive must take into account the maximum ack delay and idle timeout.

atlesn:
OK thanks, I'll have a look at the max ACK delay parameter.

tatsuhiro-t:
You should also read https://datatracker.ietf.org/doc/html/rfc9000#section-10.1.2
There is no requirement to send immediate acknowledgement for received PING frame.

atlesn:
Yes, I'll have a look into that. I've implemented nghttp2 some time ago and it always responds to PINGs. With ngtcp2 i simply need to call a write function again after max ACK delay has expired to make sure the an ACK will be sent? I guess this probably goes for other frame types than PING as well.

tatsuhiro-t:
You need to call any write function when a timer is expired.
See https://nghttp2.org/ngtcp2/programmers-guide.html#timers

atlesn:
Thanks, I've overlooked the expiry functions completely.

PR #521, opened 2022-08-20
Do not trigger congestion event when PMTUD packet is lost

tatsuhiro-t:


WesleyRosenblum:
This change causes the [QUIC interop ECN tests to fail](https://interop.seemann.io/logs/2022-08-22T16:08/ngtcp2_ngtcp2/ecn/). This is because the ECN interop tests require all packets to be marked with ECT(0) or ECT(1), and this change makes PMTUD packets not be marked as ECN capable.

To work around this, would you be able to disable MTU probing for the ECN interop test?

tatsuhiro-t:
@WesleyRosenblum https://github.com/ngtcp2/ngtcp2/pull/525 will disable PMTUD for ECN test.

PR #536, opened 2022-09-02
Do not send PMTUD packet to an unvalidated path

tatsuhiro-t:


lhuang04:
`NGTCP2_DCID_FLAG_PATH_VALIDATED` will only set for client after it receives [conn_recv_path_response](https://www.internalfb.com/code/fbsource/[c7e202ef76a7037ec904d10c6d66125cf7c77e80]/third-party/ngtcp2/lib/ngtcp2_conn.c?lines=5921%2C5958). Does that mean the client can't send PMTU unless it receives the path response?

tatsuhiro-t:
Yes, I think client also validates server address when it successfully processed handshake packets.  Will fix it.

tatsuhiro-t:
#559 fixes it.

PR #542, opened 2022-09-04
User control over PING frames

pavelxdd:
Hello. I'm writing a proxy application betweeen QUIC server and client.

I need to know when one side is receiving a PING frame to send it to the other side.

Setting keep-alive doesn't help, because it hides the control from user application over PING frames anyway.

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -662,6 +662,7 @@ int Client::init(int fd, const Address &local_addr, const Address &remote_addr,
       ngtcp2_crypto_version_negotiation_cb,
       ::recv_rx_key,
       nullptr, // recv_tx_key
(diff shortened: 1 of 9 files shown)


tatsuhiro-t:
I think they are separate connection and it is the proxy that decides it needs to make connection keep alive.
Each connection has unique connection setup (e.g., idle timeout).

pavelxdd:
I want to passthrough PING frames from one connection to another. Currently there's no way to detect and generate PING frames from the user code.

An example: the Firefox doesn't close streams when I quit the browser, so if I setup keep-alive on both connections, my proxy applications continues to send PING frames to both endpoints after closing the Firefox.

PR #544, opened 2022-09-05
Skip examples if c++ compiler does not support C++20

tatsuhiro-t:


Saigut:
Hello, You should modify the examples code to support compilers below c++20 instead of requiring the compiler to support c++20.

PR #550, opened 2022-09-07
Examples tests

icing:
Adding a pytest suite for the clients/servers built in examples. Testing:

* session-less handshake on default config
* session establishment and resume
* earlydata acceptance/rejection
* client certificates present/missing
* specific cipher configured on client side

See `examples/tests/README.rst` for more details.


examples/tests/test_02_resume.py
```diff
@@ -0,0 +1,42 @@
...
+        cr = client.http_get(server, url=f'https://{env.example_domain}/', use_session=True)
+        assert cr.returncode == 0
+        cr.assert_resume_handshake()
+        # restart the server, do it again
+        server.restart()
+        cr = client.http_get(server, url=f'https://{env.example_domain}/', use_session=True)
+        assert cr.returncode == 0
+        # some clients do not Finish the handshake when resumption was rejected
+        if client.can_resume_reject:
+            cr.assert_non_resume_handshake()
```
tatsuhiro-t: I expect that the handshake succeeds even if session ticket is unusable, so cr.assert_non_resume_handshake should pass for all combinations of client and server.  Why some of those handshakes fail?

examples/tests/ngtcp2test/client.py
```diff
@@ -0,0 +1,182 @@
...
+    def http_get(self, server: ExampleServer, url: str, extra_args: List[str] = None,
+                 use_session=False, data=None,
+                 credentials: Credentials = None,
+                 ciphers: str = None):
+        args = [
+            self.path, '--exit-on-all-streams-close',
+            f'--qlog-file={self._qlog_path}'
+        ]
+        if use_session:
+            args.append(f'--session-file={self._session_path}')
```
tatsuhiro-t, line 156: I think --tp-file should also be used here to remember QUIC transport parameters.

examples/tests/test_02_resume.py
icing: That's an excellent question, why `picotls` and `boringssl` clients fail to make the handshake. I was not able to immediately see the problem. Can have another look at it.

examples/tests/ngtcp2test/client.py
icing: Can add that.

examples/tests/test_02_resume.py
icing: Turns out, the log scanner was not robust enough when crypto hexdumps were interleaved with other dumps. Improved this and now all clients pass resume tests.

examples/tests/ngtcp2test/client.py
icing: Added.

examples/tests/test_02_resume.py
tatsuhiro-t: Great!  I also see some failed scenario in bsslclient which is fixed by https://github.com/ngtcp2/ngtcp2/commit/c309f4d29ec867a405f5363022433407a9454344

tatsuhiro-t: just resumption vs early data: currently resumption test sends early data.  To stop sending early data, and just do resumption, client --disable-early-data does the job.

icing: Added this in the test_02 runs.
Also added an extra test_03 resumption with it to see that rejection does not happen.

tatsuhiro-t:

PR #565, opened 2022-09-13
Examples Tests Improvement

icing:
- inspect only hex dumps from logs that are preceded by a line matching `r'Ordered CRYPTO data in \S+ crypto level'`, so that interleaved stream/earlydata does not interfere.

### DIFF
--- a/examples/tests/ngtcp2test/client.py
+++ b/examples/tests/ngtcp2test/client.py
@@ -32,9 +32,12 @@ class ClientRun:
     @property
     def handshake(self) -> List[HSRecord]:
         if self._data_recs is None:
-            self._data_recs = [data for data in HexDumpScanner(source=self.log_lines)]
+            crypto_line =  re.compile(r'Ordered CRYPTO data in \S+ crypto level')
+            scanner = HexDumpScanner(source=self.log_lines,
+                                     leading_regex=crypto_line)
+            self._data_recs = [data for data in scanner]
             if self.env.verbose > 1:
-                log.debug(f'detected {len(self._data_recs)} hexdumps '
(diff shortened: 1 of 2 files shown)


examples/tests/ngtcp2test/client.py
```diff
@@ -82,11 +85,11 @@ def _assert_hs(self, c_hs, s_hs):
         assert self.server.hs_stripe == s_hs, \
             f'Expected "{s_hs}", got "{self.server.hs_stripe}"\n'
 
-    def assert_non_resume_handshake(self):
+    def assert_non_resume_handshake(self, allow_hello_retry=True):
```
tatsuhiro-t, line 88: Is allow_hello_retry a new parameter?  It looks like it is not used anywhere and always True.

icing: I used it to test the recent improvements in wolfSSL. I think, given we configure clients/servers by default with the same ciphers/groups, the tests should verify in the future that the handshake succeeds without an additional roundtrip.

tatsuhiro-t: So this is the feature that the future improvements use to disallow HRR.

icing: The HRR case that happend (fixed now in wolfssl master), was between boringssl server and wolfssl client. wolfssl sent a KEY_SHARE with a group that was not the highest in SUPPORTED_GROUPS. boringssl only looked for the highest group match, found no key share and triggered HRR.

tatsuhiro-t: Yeah, this HRR scenario happens between implementations, and the thing is we do not save a group used in the initial connection.  We just workaround it to set the same group preference to all implementations.

PR #651, opened 2023-01-10
Make hard-coded static library suffix optional

donny-dont:
Add `STATIC_LIB_SUFFIX` to set the suffix directly.

This follows the same behavior as nghttp2 and nghttp3.


tatsuhiro-t:

PR #668, opened 2023-02-02
Hide ngtcp2_conn_stat from public API set

tatsuhiro-t:
ngtcp2_conn_stat includes internal data structure and not suitable for public use.  We will provide some kind of connection information like struct tcp_info at some point, but it will be something different from ngtcp2_conn_stat.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -114,7 +114,8 @@ HFILES = \
 	ngtcp2_objalloc.h \
 	ngtcp2_rcvry.h \
 	ngtcp2_net.h \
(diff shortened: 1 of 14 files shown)


amirlivneh:
@tatsuhiro-t, we currently rely on `ngtcp2_conn_stat` for telemetry, e.g. to monitor RTT and CWND in production. Do you expect the API replacing it to be added soon?

tatsuhiro-t:
No ETA is set at the moment.
I'll revert this PR for now.  This will be redone when a replacement is implemented.

amirlivneh:
Thank you.

PR #671, opened 2023-02-05
Hide ngtcp2 conn stat take2

tatsuhiro-t:


tatsuhiro-t:
@amirlivneh I added ngtcp2_conn_info which is currently a subset of ngtcp2_conn_stat.  Does ngtcp2_conn_info satisfy your use case?

amirlivneh:
@tatsuhiro-t, it does. Thank you.

PR #677, opened 2023-02-19
Adjust pacing timer if connection waited too long because of overhead

tatsuhiro-t:


lib/ngtcp2_conn.c
```diff
@@ -2497,8 +2497,15 @@ static void conn_cancel_expired_pkt_tx_timer(ngtcp2_conn *conn,
...
 
 static int conn_pacing_pkt_tx_allowed(ngtcp2_conn *conn, ngtcp2_tstamp ts) {
-  return conn->tx.pacing.next_ts == UINT64_MAX ||
-         conn->tx.pacing.next_ts <= ts + NGTCP2_PKT_PACING_OVERHEAD;
+  if (conn->tx.pacing.next_ts == UINT64_MAX) {
+    return 1;
+  }
+
+  if (conn->tx.pacing.next_ts > ts) {
+    conn->tx.pacing.surplus += conn->tx.pacing.next_ts - ts;
```
amirlivneh: It's not obvious from the function name that it has a side effect, and that the frequency at which it is called can affect the pacing rate.

tatsuhiro-t: Right, I should find another stable point to do this.

tatsuhiro-t: b9743f5 fixes this.  The calculation of excess was also wrong.  In theory, this patch can improve performance if a connection waited too long because there are too many operations/events in between, but so far I could not verify this.

amirlivneh: If you merge the change with a temporary setting for disabling it, we can test its impact on our end, across a wide range of mobile devices and networks.

tatsuhiro-t: Thank you.  I will add a flag to enable this to ngtcp2_settings.

tatsuhiro-t: Added ngtcp2_settings.adjust_pacing_timeout to enable this function.

amirlivneh: Update: We haven't seen any impact when Cubic is used. We're repeating the experiment with BBRv2, where initial results show some latency improvements.

tatsuhiro-t: Thank you for updates.  That makes sense because bbr actively adjusts pacing period to find bottleneck bandwidth whereas we just spread cwnd over rtt for cubic.

amirlivneh: The experiment with BBRv2 in our Android app showed a very small and barely statistically significant improvement in latency of upload requests. We don't see any effect on other requests, e.g. API requests with body. The difference is too small to have a measurable impact on user experience. At least for our use case, the extra complexity introduced by this change is probably not justified.

tatsuhiro-t: Thank you for the measurements.  That helps me a lot.
Because this change does not make any measurable improvements, I will revert this PR.

PR #686, opened 2023-02-23
Make the key update feature be supported by QUIC v2.

haproxyFred:
There was only "quicv2 ku" HKDF label which was missing.


tatsuhiro-t:

PR #692, opened 2023-03-03
Clang has __popcnt for ARM

zcbenz:
When building on Windows with clang for ARM targets, `__popcnt` is defined.


tatsuhiro-t:

PR #694, opened 2023-03-05
Do no start wev watcher when pacing timer is set

### LINKED ISSUE #693: ngtcp2_conn_get_send_quantum() usage
The programmer's guide says:

> After one or more calls of ngtcp2_conn_writev_stream() (it can be called multiple times to fill the buffer sized up to ngtcp2_conn_get_send_quantum() bytes), call ngtcp2_conn_update_pkt_tx_time() to set the timer when the next packet should be sent. The timer is integrated into ngtcp2_conn_get_expiry().

But after sending a quantum, client.cc not only waits for the timer to send the next packet, but also starts a socket writable watcher:
https://github.com/ngtcp2/ngtcp2/blob/694f3c4afc7d7b11161e003265e30420eba7f37a/examples/client.cc#L1086-L1090

Which one is the correct behavior?

tatsuhiro-t:
Do no start wev watcher when pacing timer is set.

There are still 2 cases that wev watcher is explicitly enabled:

- when signaling write event
- UDP write is blocked

Fixes #693

### DIFF
--- a/examples/client.cc
+++ b/examples/client.cc
@@ -949,6 +949,8 @@ int Client::on_write() {
     if (tx_.send_blocked) {
       return 0;
     }
+
+    ev_io_stop(loop_, &wev_);
   }
 
   if (auto rv = write_streams(); rv != 0) {
@@ -1060,7 +1062,6 @@ int Client::write_streams() {
     if (nwrite == 0) {
       // We are congestion limited.
       ngtcp2_conn_update_pkt_tx_time(conn_, ts);
-      ev_io_stop(loop_, &wev_);
       return 0;
     }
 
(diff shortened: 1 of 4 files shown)


amirlivneh:
I think this line may be redundant now:

https://github.com/ngtcp2/ngtcp2/blob/2f1509ea468bbdfdcd7849ed42d9d51e100d9309/examples/client.cc#L1531

If `tx_.send_blocked == true` and `send_blocked_packet()` was called then `wev_` is necessarily already started.

tatsuhiro-t:
Thank you for pointing this out.  #704 fixes it.

PR #775, opened 2023-05-08
ngtcp2 client experiments with no http3 dependency

iguberman:
1. Simplified examples with only OpenSSL and Boringssl client and no `nghttp3`  dependency (complies but doesn't connect to server). 
2. . `simpleclient` with SSLtrace enabled -> fails handshake


iguberman:
Created by mistake.

PR #778, opened 2023-05-15
Free objalloc in case of acks init failure

hyunjic:


tatsuhiro-t:

tatsuhiro-t:
It looks like this is not needed, because we have nothing allocated in ngtcp2_objalloc_init.

tatsuhiro-t:
Right, we do similar thing in conn_new.  Keep this for consistency.

PR #825, opened 2023-06-04
Fix timestamps during server initiated key update

NishantNori:
The timestamps are not set to the then current time when ngtcp2 API calls are made from write_streams() in examples/server.cc. Instead, all the ngtcp2 API calls use the 'ts' parameter that is set at the beginning of the function.

ngtcp2 has conn_update_timestamp() function that updates the latest timestamp and asserts the monotonicity i.e. that the newer timestamp is later than the older timestamp.

With [PR#796](https://github.com/ngtcp2/ngtcp2/pull/796) ngtcp2 asserts the monotonicity of the timestamps for server initiated key updates also.

The server initiated key update can be invoked during the lifetime of the write_streams() function and when it is invoked, the then current timestamp is submitted to the ngtcp2 API. So, the ngtcp2 latest timestamp is updated to a newer value.

When further code is executed in write_streams() function and ngtcp2 API is called with the old timestamp (set at the beginning of the function), we have an assertion failure in conn_update_timestamp().

In order to fix this, with this PR, we pass the latest timestamp always in the write_streams() function in server.cc (except when ngtcp2_conn_writev_stream() returns NGTCP2_ERR_WRITE_MORE as per the [documentation](https://github.com/ngtcp2/ngtcp2/blob/main/lib/includes/ngtcp2/ngtcp2.h#L4462)).

### DIFF
--- a/examples/server.cc
+++ b/examples/server.cc
@@ -1679,6 +1679,8 @@ int Handler::write_streams() {
 
   max_pktcnt = std::min(max_pktcnt, static_cast<size_t>(config.max_gso_dgrams));
 
+  ssize_t nwrite = 0;
+
   for (;;) {
     int64_t stream_id = -1;
     int fin = 0;
@@ -1706,7 +1708,11 @@ int Handler::write_streams() {
       flags |= NGTCP2_WRITE_STREAM_FLAG_FIN;
     }
 
-    auto nwrite = ngtcp2_conn_writev_stream(
+    if (nwrite != NGTCP2_ERR_WRITE_MORE) {
+      ts = util::timestamp();
+    }
+
+    nwrite = ngtcp2_conn_writev_stream(
         conn_, &ps.path, &pi, bufpos, max_udp_payload_size, &ndatalen, flags,
         stream_id, reinterpret_cast<const ngtcp2_vec *>(v), vcnt, ts);
     if (nwrite < 0) {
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for reporting this bug.  I think setting same ts when doing GSO is a good thing because internally it is considered as a single operation happens at the same time.  I will revert the part of #796 that updates timestamp in ngtcp2_conn_initiate_key_update.  Also will make sure that this kind of thing does not happen in the other places.

tatsuhiro-t:
If I understand it correctly, examples/server does not call ngtcp2_conn_initiate_key_update at all.  This means we do not need this patch.
If someone ever changes the server to call the function between the calls of ngtcp2_conn_writev_stream for their own needs, it is their own issue.
You can arrange the code so that the function is called before or after write_streams.

NishantNori:
Thanks for taking a look at the PR. 

If server.initiate_key_update is set to true, then ngtcp2_conn_initiate_key_update is called (from the update_key callback) between two calls of ngtcp2_conn_writev_stream from write_streams in examples/server.cc.

tatsuhiro-t:
update_key callback should not call ngtcp2_conn_initiate_key_update.  ngtcp2_crypto_update_key_cb does not do that.
update_key callback is called in response to the changing key by a remote endpoint.  It is not a callback that a endpoint initiates key update.

tatsuhiro-t:
conn_initiate_key_update might be called from callback chain, but it does not update timestamp.
It is ngtcp2_conn_initiate_key_update that update timestamp, and it is not called from callback chain.

NishantNori:
Thank you for your response. 

I took a closer look at it and it is an issue in our ngtcp2 repo due to a local patch. This fix is not needed. Closing the PR.

PR #830, opened 2023-06-07
keep-alive timeout defaults to UINT64_MAX

tatsuhiro-t:
This changes the default value of keep-alive timeout to UINT64_MAX. Previously, it is 0, and it is the value that disables keep-alive. Now, to disable keep-alive, set it to UINT64_MAX.


JunqiWang:
Hi, can you share what is the motivation of this change?

PR #919, opened 2023-08-29
Make crypto static library targets explicitly STATIC

jagerman:
Not having the STATIC keyword in `add_library(...)` makes CMake go look for a `BUILD_SHARED_LIBS` variable and, if set, makes the "static" library crypto targets actually create a shared library.

This causes a problem if including ngtcp2 as a subproject from some other project that itself happens to have a default of `BUILD_SHARED_LIBS=ON`, resulting in the "static" targets getting built dynamically.

This fixes it by making the crypto static library targets explicitly STATIC so that this cmake messiness doesn't happen (and also matches the way the main ngtcp2_static target is created).

### DIFF
--- a/crypto/boringssl/CMakeLists.txt
+++ b/crypto/boringssl/CMakeLists.txt
@@ -45,7 +45,7 @@ endforeach()
 
 if(ENABLE_STATIC_LIB)
   # Public static library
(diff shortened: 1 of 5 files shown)


tatsuhiro-t:

PR #958, opened 2023-09-27
Fix missing prefix for AF_INET macros in ngtcp2_crypto.c

jpboivin:
This a fix we had on our side. The remaining of the lib is using `NGTCP2_` prefixed `AF_INET` macros, but this file was not for these two lines.

FYI: We didn't update samples/tests to use the ngtcp2 macros.

### DIFF
--- a/crypto/shared.c
+++ b/crypto/shared.c
@@ -1086,11 +1086,11 @@ static size_t crypto_generate_regular_token_aad(uint8_t *dest,
   size_t addrlen;
 
   switch (sa->sa_family) {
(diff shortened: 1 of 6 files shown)


tatsuhiro-t:
Thank you for PR.  I think at least source files under tests directory should be fixed as well to pass CI.

tatsuhiro-t:
Source files in samples should not be touched because they are not compiled if generic sockaddr is used.

tatsuhiro-t:

PR #987, opened 2023-10-23
cmake: re-add two missing includes

### LINKED ISSUE #979: build failed due to CheckCCompilerFlag module is not included
`check_c_compiler_flag` is used but `CheckCCompilerFlag` module is not included.

vszakats:
Previously these were were included transitively via `cmake/ExtractValidFlags.cmake`. #964 deleted that file. The replacements it added pulled these includes after the local code uses them when `ENABLE_ASAN` enabled.

Re-add these includes into the `CMakeLists.txt` that uses them.

Reported-by: fpliu
Fixes #979

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,9 @@ foreach(_build_type "Release" "MinSizeRel" "RelWithDebInfo")
   endforeach()
 endforeach()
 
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:

PR #1006, opened 2023-11-03
Use __FILE_NAME__ instead of __FILE__

### LINKED ISSUE #1003: Absolute source paths leaking into the release binaries
Since 31fa4fe7446a8456fa1cca4c579334861fd5a069 #585, the
`ngtcp2_unreachable()` macro leaks the full path of the source files
it's used in, when building with CMake. (Haven't tried autotools, which
is less prone to referencing sources with their full path, but may be
affected nevertheless.)

The full path makes it difficult to make reproducible builds accross
environments and may also leak sensitive data.

IMO one possible fix is to reduce the macro to a generic message
for non-debug builds, or replace it with a no-op. The latter would also 
avoid using `abort()` in the library code, which may badly interfere
with apps using it (Ref:
https://infosec.exchange/@harrysintonen/111325261101468978).

The same applies to the nghttp3 library:
https://github.com/ngtcp2/nghttp3/commit/e9531116d8396834c5ec55ac8254013643d11610
https://github.com/ngtcp2/nghttp3/pull/93

tatsuhiro-t:
Use __FILE_NAME__ instead of __FILE__ not to include a full source file path.

Fixes #1003

### DIFF
--- a/lib/ngtcp2_unreachable.h
+++ b/lib/ngtcp2_unreachable.h
@@ -31,8 +31,14 @@
 
 #include <ngtcp2/ngtcp2.h>
 
+#ifdef __FILE_NAME__
+#  define NGTCP2_FILE_NAME __FILE_NAME__
+#else /* !__FILE_NAME__ */
+#  define NGTCP2_FILE_NAME "(file)"
+#endif /* !__FILE_NAME__ */
+
 #define ngtcp2_unreachable()                                                   \
-  ngtcp2_unreachable_fail(__FILE__, __LINE__, __func__)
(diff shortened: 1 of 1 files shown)


vszakats:
LGTM. Also see https://github.com/ngtcp2/ngtcp2/issues/1003#issuecomment-1791849592 about a suggestion to shift to unique symbol names, that might get the project closer to unity/jumbo builds with CMake.

tatsuhiro-t:
As for unity build, currently I have no plan for that.  I prefer allowing same static function names in individual files to gaining a bit of speed just one time compilation.

PR #1017, opened 2023-11-22
Handle unsigned conversion warning

hyunjic:
With recent change to the type of frame type, this part gives me conversion warning when running unit tests.

```
unsigned conversion from ‘int’ to ‘uint64_t’ {aka ‘long unsigned int’} changes value from ‘-8’ to ‘18446744073709551608’
[-Wsign-conversion]
     if ((type & ~(NGTCP2_FRAME_STREAM - 1)) == NGTCP2_FRAME_STREAM) {
```

### DIFF
--- a/lib/ngtcp2_pkt.c
+++ b/lib/ngtcp2_pkt.c
@@ -539,7 +539,8 @@ ngtcp2_ssize ngtcp2_pkt_decode_frame(ngtcp2_frame *dest, const uint8_t *payload,
     return ngtcp2_pkt_decode_datagram_frame(&dest->datagram, payload,
                                             payloadlen);
   default:
-    if ((type & ~(NGTCP2_FRAME_STREAM - 1)) == NGTCP2_FRAME_STREAM) {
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.
type is of uint8_t.  Where uint64_t came from?

hyunjic:
It turned out that the warning was introduced by our local change to support frame types > 0xff made [before the recent type change](https://github.com/ngtcp2/ngtcp2/commit/c503b41fb5534bb606cc3d93297dd07ec8c4f43c). We changed the local type variable to uint64_t. I will close the PR.

Out of curiousity, are you keeping uint8_t in decode function for performance reason?

tatsuhiro-t:
I am not sure how much performance we gain from uint8_t, but I keep it as is because we just check one byte for the standard types that are explicitly allowed to close a connection if they are not encoded in 1 byte.  We have no reason to accept those that encode them in multiples bytes.
For the future extension frames, use new 64 bits frame type variable when needed as I just wrote in a source code comment.

PR #1021, opened 2023-11-25
autotools: Check wolfSSL_is_quic is available

tatsuhiro-t:


mikeswhite1:
I'm not certain what's going on yet, but after this commit ngtcp2 is failing to configure for with the same wolfSSL build I've been using for a bit now~

> checking for WOLFSSL... yes
> checking for wolfSSL_is_quic... no
> configure: wolfSSL does not have QUIC interface, disabling it
> configure: error: wolfssl was requested (--with-wolfssl) but not found

(wolfssl build configured with **--enable-quic**)

mikeswhite1:
Oh I think it is because the script is passing a NULL pointer to **WOLFSSL_IS_QUIC** which returns false in this case

from wolfssl:
`#define WOLFSSL_IS_QUIC(s)  (((s) != NULL) && ((s)->quic.method != NULL))
`

the configuration script in ngtcp2:
```
WOLFSSL *ssl = NULL;
wolfSSL_is_quic(ssl);
```

tatsuhiro-t:
Check config.log to see what is really going on.

mikeswhite1:
Found out why. If you pass on an environment variable for the path to the library it does not get evaluated properly in the configuration script -- not sure if that's something that is fixable or my fault for trying

ie you can't do this anymore:
`--with-wolfssl WOLFSSL_CFLAGS='-I${PATH}/include'`

tatsuhiro-t:
Right, so this is  user error, not ngtcp2 bug.

PR #1063, opened 2024-01-14
Fix incorrect iteration in loss detection

winshining:
There is no iteration when a packet is declared lost between outer **for** iteration and inner **for** iteration, which makes check in original line 1025 in ngtcp2_rtb.c:

    if (last_lost_pkt_num == ent->hd.pkt_num + 1 && ent->ts >= start_ts)

will never be met even if the packet number after the lost one is consecutive.


tatsuhiro-t:
Thank you for reporting this.  I open PR #1064 to fix this differently in a way to reduce code duplicates.

tatsuhiro-t:
Closing in favor of #1064

PR #1111, opened 2024-02-10
examples: Add include in GnuTLS example

atlesn:
Added include of <algorithm> as this seems to be required when building on recent Arch linux against GCC libc++.

The compiler was complaining about `std::copy_n` not being defined.


tatsuhiro-t:

PR #1128, opened 2024-02-19
Make Path MTU Discovery probes configurable

### LINKED ISSUE #1114: Make PMTUD probes configurable
Would it be possible to make the `mtu_probes` configurable, e.g., via `ngtcp2_settings`?

One main reason would be so one could use `- 28` instead of `- 48` if they know the QUIC connection will be over IPv4 not IPv6.

Another reason would be to include `1500 - 28` for a very common use case (just regular ethernet with no tunnel).

Finally, one might want to include `9000 - 28` (or `9000 - 48`) if there is a chance all nodes on the network could support jumbo frames.

Thanks!

tatsuhiro-t:
Fixes #1114


sshock:
Nice work, thanks!

PR #1169, opened 2024-03-17
Add padding to at most 1200 bytes

tatsuhiro-t:


sshock:
Hi @tatsuhiro-t.  I'm curious why this change was made?

I understand padding to at most 1200 is desired when performing PMTUD, as the larger probes come after, but when not doing PMTUD isn't it better to pad initial packets to the `max_tx_udp_payload_size`, so that if incorrectly specified to something larger than the network can actually handle, the problem will be discovered immediately with a connection failure rather than later on at some random point when a large packet gets dropped?

tatsuhiro-t:
See #1801 that pads to max_tx_udp_payload_size if ngtcp2_settings.no_tx_udp_size_shaping is set to nonzero.

sshock:
> See #1801 that pads to max_tx_udp_payload_size if ngtcp2_settings.no_tx_udp_size_shaping is set to nonzero.

Nice, thanks!

PR #1204, opened 2024-04-09
chore: Build multi-platform Docker image during CI

larseggert:
See https://docs.docker.com/build/ci/github-actions/multi-platform/

### DIFF
--- a/.github/workflows/interop-build.yml
+++ b/.github/workflows/interop-build.yml
@@ -15,10 +15,13 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  It looks like Dockerfile itself needs some adjustments.  Currently it does not work with linux/arm64.
Also I noticed linux/arm64 build is unbearably slow.  Will look into it.

tatsuhiro-t:
I managed to workaround the super-slow build by cross compiling everything and stitch artifacts together in final step on target image.

tatsuhiro-t:

PR #1251, opened 2024-06-16
Fix examples for picotls + libressl

nak3:
As the variables are excluded for LibreSSL in picotls's [openssl.h](https://github.com/h2o/picotls/blob/master/include/picotls/openssl.h), this patch excludes it by `#if PTLS_OPENSSL_HAVE_*`.
Without this patch, it is not possible to build the examples for picoTLS + libressl.


tatsuhiro-t:

PR #1269, opened 2024-07-20
Add #define WOLFSSL_QUIC for access wolfssl quic defines

talregev:
Add #define WOLFSSL_QUIC for access wolfssl quic defines

https://github.com/wolfSSL/wolfssl/blob/master/wolfssl/quic.h#L33


tatsuhiro-t:
wolfssl/options.h should define WOLFSSL_QUIC otherwise the wolfssl installation lacks QUIC capability.

talregev:
> wolfssl/options.h should define WOLFSSL_QUIC otherwise the wolfssl installation lacks QUIC capability.

Yes I understand. I will need to add quic feature to wolfssl. 
thank you for the hint.

PR #1270, opened 2024-07-20
examples: use Linux extensions only where available

pinotree:
Use certain Linux extensions for sockets only in case they are available, i.e. on Linux or on an OS that implements them.


tatsuhiro-t:
Could you do the same changes for examples/h09server.cc as well?

pinotree:
> Could you do the same changes for examples/h09server.cc as well?

Sure thing, done :)

tatsuhiro-t:
Thank you.  Comparing these files diff suggests that all these macros are not guarded properly.
For example:
- https://github.com/ngtcp2/ngtcp2/blob/7708062f7aef22f85cef5a9e7029ff821dc379c6/examples/h09server.cc#L2390
- https://github.com/ngtcp2/ngtcp2/blob/7708062f7aef22f85cef5a9e7029ff821dc379c6/examples/server.cc#L2378

tatsuhiro-t:
Rethinking about this change, in general it is nice to guard these extensions with macros if they are optional feature.
But they are required for these programs.  Without them, they do not work, just dropping all incoming packets on the floor.

Issue #935, opened 2023-09-06
compiler warnings with LibreSSL 3.7.3


vszakats:
First warning is because `SSL_CIPHER_get_id` returns an `unsigned long` with LibreSSL (also in latest 3.8.1) and `uint32_t` with OpenSSL/quictls (latest 3.1.2) and BoringSSL.

The second one is due to an extra function pointer `set_read_secret`. This is missing from OpenSSL/quictls (latest 3.1.2), but present in LibreSSL (also in latest 3.8.1) and BoringSSL.

```
Linux llvm 16 and Apple clang 14:
ngtcp2/crypto/quictls/quictls.c:195:15: warning: implicit conversion loses integer precision: 'unsigned long' to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
ngtcp2/crypto/quictls/quictls.c:824:1: warning: missing field 'set_read_secret' initializer [-Wmissing-field-initializers]

mingw-64 llvm 16:
ngtcp2/crypto/quictls/quictls.c:824:1: warning: missing field 'set_read_secret' initializer [-Wmissing-field-initializers]

gcc 9:
ngtcp2/crypto/quictls/quictls.c:195:15: warning: conversion from 'long unsigned int' to 'uint32_t' {aka 'unsigned int'} may change value [-Wconversion]
ngtcp2/crypto/quictls/quictls.c:824:1: warning: missing initializer for field 'set_read_secret' of 'SSL_QUIC_METHOD' {aka 'struct ssl_quic_method_st'} [-Wmissing-field-initializers]
```

tatsuhiro-t:
LibreSSL is not officially supported by ngtcp2.  It looks like LibreSSL implements BoringSSL style callbacks that quictls has not followed yet.

vszakats:
I understand, LibreSSL 3.7.0 was the first release with QUIC support. It'd be nice to see it amongst the supported backends with ngtcp2.

I've been doing test builds with this combination for a while now and aside these warnings, builds are fine and basic HTTP/3 tests do work.

You can find test binaries here with curl, HTTP/3 and LibreSSL 3.7.3:
https://github.com/curl/curl-for-win/actions/runs/6098450816
(except bullseye, where CMake is old for ngtcp2)

tatsuhiro-t:
I do not know the actual implementation in LibreSSL, but not specifying set_read_secret should fail handshake.
[...]

PR #1284, opened 2024-08-03
Add LibreSSL and Picotls+LibreSSL to CI

nak3:
As current CI does not run with LibreSSL, some changes introduces
build failure for LibreSSL and need some fixes like bd5368b.

Hence, this patch adds LibreSSL and Picotls+LibreSSL to CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,8 +9,9 @@ env:
   OPENSSL3_VERSION: 3.1.5+quic
   BORINGSSL_VERSION: a220a6024f66c123019b5c080f6bd8bcaf75448c
   AWSLC_VERSION: v1.29.0
-  PICOTLS_VERSION: 096fc5c2ab4db1c4e0adcfdd4e75b8ee2dcc7c99
+  PICOTLS_VERSION: e4f0a27ebd1c07ebed68674258da9556fb92b46b
   WOLFSSL_VERSION: v5.7.0-stable
+  LIBRESSL_VERSION: v3.9.2
   NGHTTP3_VERSION: head
 
 jobs:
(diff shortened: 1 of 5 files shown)


nak3:
As per https://github.com/ngtcp2/ngtcp2/issues/935#issuecomment-1708563444, it might be better to have its own separated library to avoid a lot of ifdef. (I agree with that.)
However, this PR didn't create the separated library to make change small as much as possible.

Also, please be aware that this PR bumped Picotls version to include the following commits:

* https://github.com/h2o/picotls/commit/a8d42c08b6c3744a248e043abd837ba216f2444a
* https://github.com/h2o/picotls/commit/cae16c843588c836656f5c07f1a4f25323c885dc

.github/workflows/build.yml
```diff
@@ -353,18 +412,18 @@ jobs:
...
-        else
+        elif [ "${{ matrix.tls }}" = "group-a" ]; then
           PICOTLS_PREFIX="$PWD/picotls-openssl3"
+        else
+          PICOTLS_PREFIX="$PWD/picotls-libressl"
         fi
-
         PICOTLS_CFLAGS="-I$PICOTLS_PREFIX/include/"
         PICOTLS_LIBS="-L$PICOTLS_PREFIX/build -lpicotls-openssl -lpicotls-core"
-
```
tatsuhiro-t: Please do not remove empty lines unless it is absolutely needed for this change.

ci/build_libressl.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+# build patched libressl (for GitHub workflow)
+
+git clone --depth 1 -b "${LIBRESSL_VERSION}" https://github.com/libressl/portable.git libressl
+cd libressl
+mkdir build
```
tatsuhiro-t: There is no need to create build directory explicitly here.

ci/build_libressl.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+# build patched libressl (for GitHub workflow)
```
tatsuhiro-t: I think we use official libressl and not patched one.

tatsuhiro-t (changes requested):

nak3:
Thank you so much! Updated.

tatsuhiro-t:
LGTM.  Thank you.  Could you squash commits into one?

tatsuhiro-t (approved):

tatsuhiro-t:

PR #1292, opened 2024-08-15
Use OSSL_ENCRYPTION_LEVEL instead of ssl_encryption_level_t.

moritzbuhl:
Everywhere else in the file it is OSSL_ENCRYPTION_LEVEL.


tatsuhiro-t:

PR #1312, opened 2024-08-28
Fixing cmake export

talregev:
Fixing cmake export


tatsuhiro-t:
Could you explain what kind of issue this change resolves?

talregev:
This when you use ngtcp2 as a third party and you need the cmake export to read the correct targets.
I contribute to vcpkg, and we use it there to detect other ports (libraries) like that. 

I also create a PR in vcpkg:
https://github.com/microsoft/vcpkg/pull/40665

tatsuhiro-t:
cc @vszakats

lib/CMakeLists.txt
```diff
@@ -141,5 +141,11 @@ if(ENABLE_STATIC_LIB)
     DESTINATION "${CMAKE_INSTALL_LIBDIR}")
 endif()
 
+install(
+  EXPORT ${NGTCP2_TARGETS_EXPORT_NAME}
+  NAMESPACE ${NGTCP2_NAMESPACE}
+  DESTINATION ${NGTCP2_CONFIG_INSTALL_DIR}
+)
```
vszakats: Isn't this already present around line ~90 already?

talregev: No, it different. 

https://stackoverflow.com/questions/71109431/cmake-install-an-export-file-in-debug-release-alternatively
https://stackoverflow.com/questions/45777734/install-export-requires-target-from-subproject

tatsuhiro-t: Good point.  I missed that.  @talregev could you check that these lines of code?

talregev: I already test it on vcpkg and it working.

talregev: @tatsuhiro-t See the answer below.

vszakats: The two snippets _are_ the same (except formatting), so perhaps the location of this instruction is what fixed it for you in vcpkg?

Is such case, what about moving the existing instruction to this new position (keeping formatting), instead of adding a second copy?

talregev: I can try to move lines 90 to the new location and see if it solve it for vcpkg.

talregev: Change it.

vszakats (approved):
LGTM, with a suggestion to update PR title and description to reflect that this is fixing EXPORT rather than adding it.

talregev:
@vszakats Another discussion.
In vcpkg, we install the the cmake configs to "share/<port>" and in here you install it to "lib/cmake/<port>"
Can I change it in other PR to align with vcpkg?

vszakats:
@talregev: Would this change solve this problem?:
```diff
-set(NGTCP2_CONFIG_INSTALL_DIR "lib/cmake/${PROJECT_NAME}")
+set(NGTCP2_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
```

vszakats:
> > @talregev: Would this change solve this problem?:
> > ```diff
> > -set(NGTCP2_CONFIG_INSTALL_DIR "lib/cmake/${PROJECT_NAME}")
> > +set(NGTCP2_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
> > ```
> 
> I think not, because the ${CMAKE_INSTALL_LIBDIR} is still lib on vcpkg and here. Also even if it was different. vcpkg install to share/port, and not share/cmake/port.
> 
> We can take this on other PR. I am waiting for this PR will be merge. Then I will open a new PR, and I will cc you both, and we can discuss it there.

AFAICS other projects also put this file under `<lib>/cmake/` as a matter of convention.
If this is a special need of vcpkg, and I think it's best solved withing vcpkg.

talregev:
> > > @talregev: Would this change solve this problem?:
> > > ```diff
> > > -set(NGTCP2_CONFIG_INSTALL_DIR "lib/cmake/${PROJECT_NAME}")
> > > +set(NGTCP2_CONFIG_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
> > > ```
> > 
> > 
> > I think not, because the ${CMAKE_INSTALL_LIBDIR} is still lib on vcpkg and here. Also even if it was different. vcpkg install to share/port, and not share/cmake/port.
> > We can take this on other PR. I am waiting for this PR will be merge. Then I will open a new PR, and I will cc you both, and we can discuss it there.
> 
> AFAICS other projects also put this file under `<lib>/cmake/` as a matter of convention. If this is a special need of vcpkg, and I think it's best solved within vcpkg.

I already patch it in vcpkg. I just want to do a discussion if I can change that. If not, that ok. Thank you for your answer.

PR #1349, opened 2024-09-16
Silence faulty 'uninitialized' GCC warning.

Dook97:
When compiling `lib/ngtcp2_acktr.c` with GCC 14.2.1 a faulty `-Wmaybe-uninitialized` appears. This PR remedies the issue.

### DIFF
--- a/lib/ngtcp2_acktr.c
+++ b/lib/ngtcp2_acktr.c
@@ -283,6 +283,7 @@ void ngtcp2_acktr_recv_ack(ngtcp2_acktr *acktr, const ngtcp2_ack *fr) {
 
   min_ack = largest_ack - (int64_t)fr->first_ack_range;
 
+  assert(ent != NULL);
   if (min_ack <= ent->pkt_num) {
     acktr_on_ack(acktr, rb, j);
     return;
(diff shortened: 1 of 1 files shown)


lib/ngtcp2_acktr.c
```diff
@@ -293,6 +294,7 @@ void ngtcp2_acktr_recv_ack(ngtcp2_acktr *acktr, const ngtcp2_ack *fr) {
     min_ack = largest_ack - (int64_t)fr->ranges[i].len;
 
     for (;;) {
+      assert(ent != NULL);
```
tatsuhiro-t, line 297: Isn't it suffice to just single assert above?
I could not reproduce this warning with gcc 14.2.0.  If this is gcc 14.2.1 specific bug, then lets wait for 14.2.2 which fixes the bug.

Dook97: `ent` is being assigned to in the loop `ent = ngtcp2_ringbuf_get(rb, j);` and gcc is unable to deduce that this function always returns non-null, hence the loop `assert`.

Dook97: > lets wait for 14.2.2 which fixes the bug

Ok, thank you for considering. Closing.

PR #1363, opened 2024-09-21
examples: Bail out with ciphers option when TLS stack ignores it

tatsuhiro-t:
Bail out if TLS backend does not allow changing ciphers and ciphers option is used.  Remove AES-128-CCM from default ciphers if libressl is used.


Karthikdasari0423:
Thanks @tatsuhiro-t

PR #1382, opened 2024-10-15
ringbuf: silence superfluous unused function warning on NDEBUG builds

Dook97:
When building non-debug builds the `ispow2` function is unused. This PR marks it as such so that superfluous compiler warnings aren't emitted.

To reproduce:

```
CFLAGS='-Wall -DNDEBUG' ./configure
make
```

You will recieve:

```
ngtcp2_ringbuf.c:34:12: warning: 'ispow2' defined but not used [-Wunused-function]
   34 | static int ispow2(size_t n) {
      |            ^~~~~~
```


tatsuhiro-t:
Thank you for PR.  I think it would be better to use #ifdef NDEBUG, we do that in several places.

Dook97:
Like this? @tatsuhiro-t

tatsuhiro-t:
No, remove __attribute stuff, and guard the definition of ispow2 by NDEBUG.

tatsuhiro-t:

PR #1385, opened 2024-10-16
Use REQUIRED when finding a package

donny-dont:
If an `ENABLE` option is `ON` and requires a package to be present it should fail the build if its not found.


donny-dont:
👋 @tatsuhiro-t I've been burned by this behavior a few times where CMake detects an OpenSSL instance that doesn't have QUIC APIs and the build fails when it reaches curl.

tatsuhiro-t:
Thank you.  Could you check and fix CI failure? https://github.com/ngtcp2/ngtcp2/actions/runs/11374405658/job/31643025756?pr=1385

donny-dont:
> Thank you. Could you check and fix CI failure? https://github.com/ngtcp2/ngtcp2/actions/runs/11374405658/job/31643025756?pr=1385

What's the behavior you want here? Should ngtcp2 build without any backend?

tatsuhiro-t:
I think setting ENABLE_OPENSSL=OFF would be reasonable approach.  Previously we are ok if openssl is not found, but now we are not, so disable what is not supposed to work.

tatsuhiro-t:
Thank you for the prompt fix.
The direction is good.  Failing script on configuration failure is what configure.ac does too.
There is one issue left.   picotls backend can be enabled with openssl but it does not require QUIC API.
That is why we define VANILLA_OPENSSL_INCLUDE_DIRS and VANILLA_OPENSSL_LIBRARIES.

What we are doing in configure.ac is that by default --with-openssl is "check", which means just check openssl availability but does not fail if it is not found or does not provide QUIC API.  Then if picotls is enabled, make sure openssl is available regardless of QUIC API.

donny-dont:
So I'm taking the route of if either `ENABLE_OPENSSL` or `ENABLE_PICOTLS` is present then it looks for an OpenSSL package. It will only check for the QUIC API if `ENABLE_OPENSSL` is present and fail if its not available. That way a vanilla OpenSSL can be used for the picotls backend.

I'm of the opinion that if you enable something explicitly it should fail if what's being enabled isn't present so you aren't surprised by the results.

Happy to change things if you want a different route taken @tatsuhiro-t . I'm not familiar with configure.ac files, I primarily develop on Windows so CMake is the way there, so I'm not sure if that would also need changes for parity if this is accepted as is.

tatsuhiro-t:
> I'm of the opinion that if you enable something explicitly it should fail if what's being enabled isn't present so you aren't surprised by the results.

Agreed.

> Happy to change things if you want a different route taken @tatsuhiro-t . I'm not familiar with configure.ac files, I primarily develop on Windows so CMake is the way there, so I'm not sure if that would also need changes for parity if this is accepted as is.

This PR closely matches what configure.ac does.  LGTM.

tatsuhiro-t:

PR #1396, opened 2024-10-26
Tweak threshold for max_stream_data and max_data transmission

tatsuhiro-t:
The previous value is too conservative and does not work well with auto-tuning.


tatsuhiro-t:
This opens up window size a bit too quickly, but it would be better than choking sender.

PR #1404, opened 2024-11-04
Add also documentation building into the release

pemensik:
Include additional files required for make html target. Documentation for the release is not a bad thing too.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,7 +46,8 @@ EXTRA_DIST = \
 	cmake/FindJemalloc.cmake \
 	cmake/PickyWarningsC.cmake \
 	cmake/PickyWarningsCXX.cmake \
-	cmake/Version.cmake
+	cmake/Version.cmake \
+	SECURITY.md
 
 # Format source files using clang-format.  Don't format source files
 # under third-party directory since we are not responsible for their
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -31,6 +31,8 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
 
+EXTRA_DIST = mkapiref.py source/index.rst source/programmers-guide.rst
+
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -31,7 +31,8 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
 
-EXTRA_DIST = mkapiref.py source/index.rst source/programmers-guide.rst
+EXTRA_DIST = mkapiref.py source/conf.py.in \
+	     source/index.rst source/programmers-guide.rst
 
 # Put it first so that "make" without argument is like "make help".
 help:


pemensik:
it seems release 1.8.1 also lacks these files. My attempt to make documentation has failed without it. That documentation seems good, I think should be part of release archive.

Makefile.am
```diff
@@ -30,7 +30,7 @@ if ENABLE_EXAMPLES
 SUBDIRS += third-party examples
 endif
 
-dist_doc_DATA = README.rst
+dist_doc_DATA = README.rst SECURITY.md
```
tatsuhiro-t: This installs SECURITY.md under $prefix/share/doc/ngtcp2.  If this file is just needed in an archive, append it to EXTRA_DIST.

doc/Makefile.am
```diff
@@ -31,6 +31,9 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
 
+dist_bin_SCRIPTS = mkapiref.py
+dist_doc_DATA = source/index.rst source/programmers-guide.rst
```
tatsuhiro-t: ditto.  If these files are just needed in an archive, add them to EXTRA_DIST.

pemensik: Hmm, while index.rst is not much useful itself, programers-guide.rst might be a good part of devel package. But I think better would be if man page were generated also. That would be a bit better consumable from terminal.

tatsuhiro-t: I have no plan to add generated manual pages to release archive at the moment.
People can access nghttp2.org/ngtcp2 to get html version of manual, or can git clone the repo.

doc/Makefile.am
```diff
@@ -61,5 +64,23 @@ cryptoapiref: mkapiref.py \
...
+
+pdf-local: apiref cryptoapiref
+	$(MAKE) $(MAKEOPTS) sphinx-build SPHINX_BUILDER=latexpdf
+
+latexpdf: pdf
+
+doc:
+	for SPHINX_BUILDER in $(SPHINX_BUILDERS); do \
+		$(MAKE) $(MAKEOPTS) sphinx-build SPHINX_BUILDER=$${SPHINX_BUILDER}; \
+	done
```
tatsuhiro-t: It would be better to run command directly rather than recursively calling make:

```make
man: apiref cryptoapiref
	@$(SPHINXBUILD) -M man "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

epub: apiref cryptoapiref
	@$(SPHINXBUILD) -M epub "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

pdf-local: apiref cryptoapiref
	@$(SPHINXBUILD) -M latexpdf "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

doc: html man epub
```

doc/Makefile.am
pemensik: I did not propose adding generated man pages. I do not think they should be part of source archive, no. But there should be a make target to create them in a simple way.

doc/Makefile.am
pemensik: Yes, it might. But that is a repeated copy & pasting rules for all specific targets, where only a minimal change occurs. I tried to make it reusing common sphinx-build target, though I am not sure how to make it nicer. If you prefer it copied separately, that is of course possible.

Makefile.am
pemensik: Okay. I think it would not hurt having this installed as part of package, but I guess security researchers can find it in source archive only.

doc/Makefile.am
tatsuhiro-t: The current proposed way is also kind of copy-and-paste.
Is there any real usecase or demand to offer this kind of extra documentation format like epub or pdf?
If this is just added for completeness or "maybe useful someday", do not add now and wait when they are actually needed.

pemensik: I added these, because ``make help`` lists them. But when they are tried, they do not work. Perhaps make help might be removed, if only make html is supported.

tatsuhiro-t: It is not relevant to this primary objective of this PR.  Please remove these targets you added.  Just add files that are needed to build doc to the release archive, and finish.  Do not do everything you find suspicious.  That would unnecessarily convolute PR.  I would like to leave help target issue now.

doc/Makefile.am
pemensik: Okay, sorry for long delays! Moved just man target adding into different PR. Modified this to add all git tracked files, which are not generated by build process. That should ``make html`` in doc directory successful even from just release tarball.

doc/Makefile.am
pemensik: Moved just man target into #1780. This should allow just make html on released archive source.

pemensik: Okay, moved it to separate PR, this should fix just make html target.

PR #1407, opened 2024-11-08
cmake: Export wolfssl crypto backend lib as CMake target

redbaron:
When consuming ngtcp make its crypto library available in `find_package(ngtcp2)`.

While at it also took opportunity to modernize CMake handling a little. More specifically rely on CMake to propagate include paths and compile definitions from dependencies. 

It also changes the way wolfssl is consumed: we are building with CMake, so expect wolfssl to provide CMake config package as well.


redbaron:
I didn't touch awkward split between shared and static lib targets as it is a bigger change, but it would be nice to
to link with ngtcp2 as `ngtcp2::ngtcp2` regardless of how it was built. Wolfssl project is a good example how CMake exporting should be done.

CMakeLists.txt
```diff
@@ -385,8 +384,23 @@ message(STATUS "summary of build options:
...
       BoringSSL:      ${HAVE_BORINGSSL} (LIBS='${BORINGSSL_LIBRARIES}')
       Picotls:        ${HAVE_PICOTLS} (LIBS='${PICOTLS_LIBRARIES}')
-      wolfSSL:        ${HAVE_WOLFSSL} (LIBS='${WOLFSSL_LIBRARIES}')
+      wolfSSL:        ${HAVE_WOLFSSL}
       Jemalloc:       ${HAVE_JEMALLOC} (LIBS='${JEMALLOC_LIBRARIES}')
       Libbrotlienc:   ${HAVE_LIBBROTLIENC} (LIBS='${LIBBROTLIENC_LIBRARIES}')
       Libbrotlidec:   ${HAVE_LIBBROTLIDEC} (LIBS='${LIBBROTLIDEC_LIBRARIES}')
 ")
+
+set(_libs_msg)
```
redbaron, line 393: this snippet attempts to reproduce previous summary of all imported libs and still work for multi config generators like `Ninja Multi-Config` where `${CMAKE_BUILD_TYPE}` var is [not available](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#build-configurations). Example output:

```
-- Dependencies:
    wolfssl[RELEASE]: /Users/<REDACTED>/build/macosx-arm64/vcpkg_installed/arm64-osx/lib/libwolfssl.a

```

tatsuhiro-t:
Could you check the CI failures?

redbaron:
CI builds ngtcp2 with CMake, but installs dependencies with autotools, denying them option to install CMake exported Config file.  Also I think it is fair to say, that dependencies are often managed by vcpkg , which also builds them with CMake. 

CMake-built dependency usuallyy install pkg-config files to be consumed by autotools, but not other way around: builds with autotools doesn't install CMake configs.  If you switch dependencies build  script to CMake and [populate](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) `CMAKE_PREFIX_PATH` for each it should work.

tatsuhiro-t:
That might brake the existing builds which depends on the current behavior.
I would be better to stick with autotools rather than making everything cmake.
@vszakats What do you think about this change?

redbaron:
Current behaviour doesn't allow consuming ngtcp2_crypto  from CMake without hacks reimplementing what CMake [export](https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html) does for free.

CMakeLists.txt
```diff
@@ -123,7 +123,7 @@ if(ENABLE_OPENSSL OR ENABLE_PICOTLS)
   find_package(OpenSSL 1.1.1 REQUIRED)
 endif()
 if(ENABLE_WOLFSSL)
-  find_package(wolfssl 5.5.0 REQUIRED)
+  find_package(wolfssl 5.5.0 CONFIG REQUIRED)
```
vszakats: wolfSSL CMake support lags behind its `./configure` AFAICT.
Does this change let the latter continue to work as it was before? If not, I'd consider leaving it as-is.

CMakeLists.txt
```diff
@@ -349,6 +346,8 @@ set(PKGDATADIR "${CMAKE_INSTALL_FULL_DATADIR}/${CMAKE_PROJECT_NAME}")
 
 install(FILES README.rst DESTINATION "${CMAKE_INSTALL_DOCDIR}")
 
+
+set(NGTCP2_TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets" )
```
vszakats:
```suggestion
set(NGTCP2_TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")

```

(In other parts of the patch, spacing is also inconsistent, including tabs.)

lib/CMakeLists.txt
```diff
@@ -86,7 +85,9 @@ include(CMakePackageConfigHelpers)
 write_basic_package_version_file(
   "${NGTCP2_VERSION_CONFIG}" VERSION ${NGTCP2_VERSION} COMPATIBILITY SameMajorVersion
 )
-configure_file("${CMAKE_CURRENT_SOURCE_DIR}/config.cmake.in" "${NGTCP2_PROJECT_CONFIG}" @ONLY)
+configure_package_config_file("${CMAKE_CURRENT_SOURCE_DIR}/config.cmake.in" "${NGTCP2_PROJECT_CONFIG}"
```
vszakats, line 88: Why remove `@ONLY`?

crypto/wolfssl/CMakeLists.txt
```diff
@@ -71,12 +65,14 @@ if(ENABLE_STATIC_LIB)
     ARCHIVE_OUTPUT_NAME ngtcp2_crypto_wolfssl${STATIC_LIB_SUFFIX}
     C_VISIBILITY_PRESET hidden
   )
-  target_compile_definitions(ngtcp2_crypto_wolfssl_static PUBLIC
-    "-DNGTCP2_STATICLIB")
```
vszakats, line 75: Why is this deleted? Is it necessary for this PR?

crypto/wolfssl/CMakeLists.txt
```diff
@@ -21,22 +21,15 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-add_definitions(-DBUILDING_NGTCP2)
```
vszakats, line 24: Why is this deleted?

CMakeLists.txt
redbaron: no it doesn't automatically falls back to pkg-config. I have pushed update to Findwolfssl module which first tries finding CMake export config from openssl and falls back to pkg-config, trying to replicate it.

lib/CMakeLists.txt
redbaron: there is no `@ONLY` [documented](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) for this function

CMakeLists.txt
vszakats: Shouldn't tweaking wolfSSL Find module be a separate PR?

Is this required to export the wolfSSL crypto lib via ngtcp2's config?

tatsuhiro-t:
If wolfssl cmake support lags behind, then there is no reason to solely rely on it.  At least, it is required to fallback to pkg-config (current way to find the lib) if cmake config is not found.

vszakats:
To my eyes the existing `Findwolfssl.cmake` is fine.

(Could be improved by using the `pkg-config` results as-is instead of a hint, but it's a deep rabbit-hole to enter, and not critical.)

crypto/wolfssl/CMakeLists.txt
redbaron: it is not required, because this definition comes with dependency on  `ngtcp2_static`

CMakeLists.txt
redbaron: > Is this required to export the wolfSSL crypto lib via ngtcp2's config?

yes it is required, because exported config needs to be useful regardless of how ngtcp2's dependencies were built.  When generating export config, CMake propagates all exported targets dependencies, so that consumers of ngtcp2_crypto_wolfssl know that wolfssl needs to be linked too. For that export to work, Findwolfssl should create targets which looks the same, regardless of how they were found.

crypto/wolfssl/CMakeLists.txt
redbaron: I incorrectly extracted cmake export part from a bigger  changset  prepared locally. reverted, but updated to a more specific `add_compile_definitions`

cmake/Findwolfssl.cmake
```diff
@@ -1,41 +1,23 @@
...
-                                  VERSION_VAR WOLFSSL_VERSION)
-
-if(WOLFSSL_FOUND)
-  set(WOLFSSL_LIBRARIES     ${WOLFSSL_LIBRARY})
-  set(WOLFSSL_INCLUDE_DIRS  ${WOLFSSL_INCLUDE_DIR})
-endif()
-
-mark_as_advanced(WOLFSSL_INCLUDE_DIR WOLFSSL_LIBRARY)
+        wolfssl_LINK_LIBRARIES wolfssl_INCLUDE_DIRS
+        VERSION_VAR WOLFSSL_VERSION)
```
vszakats: I suggest moving formatting/whitespace changes to a separate PR. It makes it easier to follow code changes.

redbaron: I went over changes and removed  unnecessary formatting changes and made it more consistent

vszakats:
I think I misunderstood the goal of this PR, I interpreted it is about exporting the ngtcp2_crypto_wolfssl lib from the cmake config, while it seems to be about doing this _and_ pulling in wolfssl as a dependency automatically. If so, I'm less sure about this. It's also mixed with not stricly related changes.

Taking a step back:

libngtcp2.pc doesn't depend on the ngtcp2_crypto_* lib (aka crypto lib). The crypto lib is exported via an independent .pc file (libngtcp2_crypto_*.pc). Which makes sense because the point of a separate crypto lib was likely that it can be built and referenced separately (even if in most cases (?) they are built in a single pass.)

This PR would break this layout for cmake and wolfssl specifically, and bundle them together in the cmake config.

I think the first question is if it's desired to make such exception. This looks inconsistent with the pkg-config way and the concept of keeping the crypto-backend specific stuff in an indepentend lib.

An alternate approach, matching the current design, would be to provide separate cmake modules for the crypto libs.

A related but not strictly necessary change in this PR is to also pull in wolfSSL itself via the cmake config file. This might be useful, though IMO it's a few steps away from where we stand now.

(For this, first the Find modules may be revised to be ready for being exported and installed, then ngtcp2 build adapted to these changes, then wire them up with the config modules where the crypto lib exported from. etc.)

redbaron:
> and pulling in wolfssl as a dependency automatically.

wolfssl as a dependency only comes when users do:

```
find_package(ngtcp2 CONFIG)
target_link_libraries(my target PRIVATE ngtcp2::ngtcp2_crypto_wolfssl_static)
```

yes, last line doesn't only link with   ngtcp2_crypto_wolfssl_static target but also all of it's dependencies, their include paths and required defined, that is with wolfssl and ngtcp2_static in this case, but if users link with `ngtcp2::ngtcp2_static` only then there will be no wolfssl required.

> This PR would break this layout for cmake and wolfssl specifically, and bundle them together in the cmake config.

They are in the same config, but separate targets, so ngtcp2 can still be consumed without ngtcp2_crypto

tatsuhiro-t:
CI error

```
[ 46%] Building C object crypto/wolfssl/CMakeFiles/ngtcp2_crypto_wolfssl.dir/__/shared.c.o
/home/runner/work/ngtcp2/ngtcp2/ngtcp2-1.9.0-DEV/crypto/shared.c:37:10: fatal error: 'ngtcp2_macro.h' file not found
   37 | #include "ngtcp2_macro.h"
      |          ^~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [crypto/wolfssl/CMakeFiles/ngtcp2_crypto_wolfssl.dir/build.make:90: crypto/wolfssl/CMakeFiles/ngtcp2_crypto_wolfssl.dir/__/shared.c.o] Error 1
```

suggests that include path is not set correctly.

redbaron:
It just occurred to me, that if wolfssl was found via pkg-config, then referencing it as a `wolfssl::wolfssl` target in exported CMake config does no good: when users consume it, no such target exists anywhere and they'd need to reimplement whole `Findwolfssl` themselves. I pushed a fix where `wolfssl::wolfssl` is listed as dependency only if it was found with CMake and therefore we can expect users to do the same reasonably easy.

PR #1443, opened 2024-12-15
Update ksl fuzzer with FuzzedDataProvider

tatsuhiro-t:


fuzz/ksl.cc
```diff
@@ -27,46 +29,49 @@ ngtcp2_ksl_search_def(less, less)
...
   ngtcp2_ksl_init(&ksl, less, ksl_less_search, keylen, ngtcp2_mem_default());
 
-  for (; size >= keylen; ++data, --size) {
-    KeyType d;
-
-    memcpy(&d, data, keylen);
+  for (; fuzzed_data_provider.remaining_bytes();) {
+    auto add = fuzzed_data_provider.ConsumeBool();
+    auto key = fuzzed_data_provider.ConsumeIntegralInRange<uint64_t>(
+      0, std::numeric_limits<uint64_t>::max());
```
amirlivneh: You can just use `ConsumeIntegral()` here.

tatsuhiro-t: Nice catch.  Will fix it.

PR #1466, opened 2025-01-03
Fix uninitialized connection close frame_type

tatsuhiro-t:


thesamesam:
Thank you! It solves the problem.

thesamesam:
By the way, would it be possible for https://github.com/nemequ/munit/pull/105 to be pulled in to the mtest fork used by your projects? It fixes the build with C23.

tatsuhiro-t:
Thank you for pointing it out.  Done https://github.com/ngtcp2/munit/commit/083b619c45aff90cf815579d83f1a9b1d5131d1d

PR #1483, opened 2025-01-23
fuzz: Exercise extending max offsets

amirlivneh:


fuzz/read_write_pkt.cc
```diff
@@ -772,6 +772,19 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
       break;
     }
 
+    if (fuzzed_data_provider.ConsumeBool()) {
+      auto stream_id = fuzzed_data_provider.ConsumeIntegral<uint64_t>();
+      auto datalen = fuzzed_data_provider.ConsumeIntegral<uint64_t>();
+
+      ngtcp2_conn_extend_max_stream_offset(conn, stream_id, datalen);
```
tatsuhiro-t: It would better to check the return value.

amirlivneh: You're right. I was confused by ngtcp2_conn_extend_max_offset() returning void.

tatsuhiro-t:

PR #1487, opened 2025-01-25
fuzz: Set negotiated version before setting remote transport params

amirlivneh:
conn_client_validate_transport_params() validates params.version_info.chosen_version by comparing it to conn.negotiated_version. To ensure this validation works correctly, conn.negotiated_version should be set before calling ngtcp2_conn_decode_and_set_remote_transport_params().


tatsuhiro-t:

PR #1550, opened 2025-02-18
Doc: fix reference to ngtcp2_conn_get_0rtt_crypto_ctx

jagerman:
The referenced `ngtcp2_conn_get_early_crypto_ctx` function doesn't exist, but `ngtcp2_conn_get_0rtt_crypto_ctx` does and appears to be what this should be referencing.


tatsuhiro-t:

PR #1567, opened 2025-02-27
Write zerolen vec

tatsuhiro-t:


redbaron:
What does it do on zero length write? Does it send STREAM frame?

PR #1578, opened 2025-03-06
Add flags to pad a packet when it is finalized

tatsuhiro-t:
NGTCP2_WRITE_STREAM_FLAG_PADDING instructs ngtcp2_conn_writev_stream to pad a packet toward the minimum size of path MTU or a given packet buffer.  It only applies to non-empty 0 RTT or 1 RTT packet.

Similary, NGTCP2_WRITE_DATAGRAM_FLAG_PADDING instructs ngtcp2_conn_writev_datagram to pad a packet.


redbaron:
Without this flag, was it limiting itself by path MTU when filling up buffer?

tatsuhiro-t:
Always limited by path MTU by default with or without the flag.

PR #1583, opened 2025-03-17
support openssl tls backend for ngtcp2

nhorman:
Add support for the openssl tls backend, making use of the newly added SSL_set_quic_tls_cbs registration function available in openssl-3.5 and later.

I've made a temporary branch https://github.com/nhorman/ngtcp2/trees/openssl-interop), based on this PR, in which I modified the ngtcp2 interop container to build with this backend, running the following quic-interop-runner tests with ngtcp2+openssl as a client and server passes the following tests:
* retry
* handshake
* multiplexing
* v2
* ecn
* resumption


tatsuhiro-t:
See #1582

nhorman:
@tatsuhiro-t  thank you.  FWIW, I haven't hit the same issues you did (more or less):
* Use of NULL BIO's - I did hit this and fixed it up in the configure_session apis for libngtcp2_crypto_openssl
* Server-side resumption and early data are unstable and broken - That doesn't seem to be the case in this implementation (based on the passing of the resumption test in quic-interop-runner).  It may have to do with the fact that I'm implemented record boundary parsing for the quic_tls_send_cb, which I had to do to ensure record boundary submission when relevant on key changes.
* CCM ciphers - Unsure about this one, but I'm happy to test.  Can I ask what test reproducer you used?
* crypto_release_rcd_cb - I never hit this problem because I never have to call SSL_get_app_data from the record release callback.

tatsuhiro-t:
I think QUIC TLS API should be tested more thoroughly by openssl dev to ensure that everything works in a way they intended to work.  Like getting NULL from SSL_get_app_data unexpectedly may lead to serious outcome.  Checking NULL even if we set something is super annoying.
Use examples client and sever, and specify CCM cipher only.
For resumption and early data, openssl never produce handshake tx key on server side.  I do not know why.

tatsuhiro-t:
Does openssl quic tls API really require that application must parse and pass data in TLS record boundary?  I fed 1 byte at a time to openssl, and it still worked.  If application ever needs to parse TLS record and just pass single TLS record to openssl, I has to say it is very hard to use.  Imagine every implementation that uses this openssl API has to parse TLS record just for the couple of messages passing during handshake.  That is awfully a lot complicated.

t8m:
@mattcaswell might want to answer some of the comments/questions by @tatsuhiro-t above.

mattcaswell:
> I think QUIC TLS API should be tested more thoroughly by openssl dev to ensure that everything works in a way they intended to work. Like getting NULL from SSL_get_app_data unexpectedly may lead to serious outcome. Checking NULL even if we set something is super annoying.
Use examples client and sever, and specify CCM cipher only.
For resumption and early data, openssl never produce handshake tx key on server side. I do not know why.

I will investigate these issues.

@tatsuhiro-t Can you expand a bit on what you see going wrong with the CCM cipher?

> Does openssl quic tls API really require that application must parse and pass data in TLS record boundary? 

It shouldn't need to. The only requirement is that record data must all be from the same epoch, e.g. you should not pass ServerHello data and EncryptedExtensions data all at the same time. But that should not happen anyway I think (since the keys to decrypt EncryptedExtensions won't have been "yielded" until after the ServerHello has been processed).

tatsuhiro-t:
>  Can you expand a bit on what you see going wrong with the CCM cipher?

I got "no cipher" error when I set CCM cipher only. quictls works with the same setup.

As for this PR, I won't merge it.  It seems that this changes contains several issues, including stack-use-after-return.  The effort to add openssl 3.5 support will be done via #1582.

t8m:
We actually do not support AES-CCM in QUIC/QUIC-TLS yet. It should be fairly easy to add it though (i.e. can be done in 3.6 as a RFE).

nhorman:
I've been able to reproduce the CCM error here.

Effectively it appears that our tls stack simply doesn't allow for CCM ciphers under quic.  From ssl_cipher_disabled:
```
if (SSL_IS_QUIC_HANDSHAKE(s))
        /* For QUIC, only allow these ciphersuites. */
        switch (SSL_CIPHER_get_id(c)) {
        case TLS1_3_CK_AES_128_GCM_SHA256:
        case TLS1_3_CK_AES_256_GCM_SHA384:
        case TLS1_3_CK_CHACHA20_POLY1305_SHA256:
            break;
        default:
            return 1;
        }
```

It appears that the inclusion of TLS1_3_CK_AES_128_CCM_8_SHA256 was intentionally omitted due to a restriction as referenced in RFC 9001, section 5.3:
```
QUIC can use any of the cipher suites defined in [TLS13] with the exception of TLS_AES_128_CCM_8_SHA256.
```

Seems like QUIC is specifically enjoined from using that cipher.

mattcaswell:
> Seems like QUIC is specifically enjoined from using that cipher.

Yes, but TLS_AES_128_CCM_SHA256 should be fine. I already have a fix locally. PR coming soon.

mattcaswell:
> > Does openssl quic tls API really require that application must parse and pass data in TLS record boundary?
> 
> It shouldn't need to. The only requirement is that record data must all be from the same epoch, e.g. you should not pass ServerHello data and EncryptedExtensions data all at the same time. 

By the way this restriction comes about because of this in RFC8446:

>       Handshake messages MUST NOT span key changes.  Implementations
>       MUST verify that all messages immediately preceding a key change
>       align with a record boundary; if not, then they MUST terminate the
>       connection with an "unexpected_message" alert.  Because the
>       ClientHello, EndOfEarlyData, ServerHello, Finished, and KeyUpdate
>       messages can immediately precede a key change, implementations
>       MUST send these messages in alignment with a record boundary.

nhorman:
FWIW, With @mattcaswell 's changes here:
https://github.com/openssl/openssl/pull/27091

This PR now works with the CCM cipher.

nhorman:
so, I managed to get the resumption issue to reproduce (not sure how it was working previously), though I'm not seeing the same behavior that you are @tatsuhiro-t .  Specifically, I'm observing that the client never actually creates a session cache file (via the new session callback).  I'm attaching the data for my observations here:
[observations.zip](https://github.com/user-attachments/files/19350676/observations.zip)


1) I'm running the following commands:
* ```opensslserver 127.0.0.1 4433 ~/git/openssl/test/certs/serverkey.pem ~/git/openssl/test/certs/servercert.pem```
* ```opensslclient --session-file ./session.txt 127.0.0.1 4433 https://localhost/index.html```

Included in the zip file are:
* test.cap - tcpdump of the capture of this connection
* keys.log - keylog file to decode the tcpdump
* output.txt - log captured from opensslclient


Notes:
1) Frame 1 contains the client hello, which corresponds to lines 1-6 in the client log, so far so good
2) Frames 2-3 contains the received server hello, which is fragmented.  This corresponds to  lines 9-137 in the client log, and appears to have been processed correctly.  
3) lines 136-149 of the client log appear to be processing the remainder of the server hello.  I note here the QUIC handshake has completed message, which makes sense.
4) lines 150-159 of the client log appear to correspond to frame 3, in which the client sends an ACK and a HANDSHAKE FINISHED message
5) lines 160-171 of the client log appear to go back to processing the first frame, which I don't understand.
6) lines 175-179 seem to be where the problem occurs.  I believe this receipt matches frame 4,in which the server sends the client new session tickets.  It appears at this point, becuase the QUIC connection has been considered established, the quic client is discarding initial packet data because it was 'delayed'.  I presume that delayed implies here that any handshake data received after the connection is established (presumably by the receiption of the above handshake finished message), is discarded as no longer being needed.  however because this is occurring the new session tickets are never making it into the tls stack, and so the new_session callback is never getting made

Question for @tatsuhiro-t: Is the expectation of the ngtcp2 quic stack that handshake data after a connection establishment should always be discarded?

Question for @mattcaswell :  Since you may just know off the top of your head, Can you confirm that it is in spec to send additional handshake data in TLS1.3 after the transmission of a handshake finished message?  I presume it is, as this works in our own non-quic clients/servers without issue (confirmed with a separate tcpdump, indicating the receipt of a new session ticket after a handshake done message).

If the answer to my question to @mattcaswell is yes, then I think we need to find a way to continue to push handshake data into the TLS stack even after the connection is established.

tatsuhiro-t:
@nhorman 

The log is very weird, and there is something very wrong in your implementation, most likely keys are not install correctly.

> Is the expectation of the ngtcp2 quic stack that handshake data after a connection establishment should always be discarded?

ngtcp2 client discards handshake packet number space when HANDSHAKE_DONE is received.

nhorman:
I believe I have the key mapping issue sorted out now, still seeing the same session cache behavior in which new session tickets aren't getting parsed.  Looking into that now.

nhorman:
ok, I've updated this branch, and figured out my weird logs.  turns out I did have the mapping for TLS encryption level to ngtcp2 encryption level wrong.  With the most recent changes, the logs look much better.

But I'm still seeing this issue where the server generates a new session ticket (observed via breakpoint in tls_construct_new_session_ticket), and it gets submitted to the quic stack via ngtcp2_conn_submit_crypto_data, but it never gets transmitted on the wire.  Looking into why that might be.

nhorman:
Ok, with the latest updates, resumption appears to be working as well

nhorman:
running this code through the quic-interop-runner shows that we're passing almost all tests here:
```
+---------+-----------------------------------------------------+----------------------------------+
|         |                        ngtcp2                       |             openssl              |
+---------+-----------------------------------------------------+----------------------------------+
|  ngtcp2 | ✓(H,DC,LR,C20,M,S,R,3,B,U,E,L1,L2,C1,C2,6,V2,BP,BA) | ✓(H,DC,LR,C20,M,S,R,3,B,L2,C2,6) |
|         |                      ?(Z,A,CM)                      |   ?(Z,U,E,A,L1,C1,V2,BP,BA,CM)   |
|         |                         ✕()                         |               ✕()                |
+---------+-----------------------------------------------------+----------------------------------+
| openssl |            ✓(H,DC,C20,M,R,3,B,A,L2,C2,6)            | ✓(H,DC,C20,M,S,R,3,B,A,L2,C2,6)  |
|         |            ?(LR,Z,U,E,L1,C1,V2,BP,BA,CM)            |  ?(LR,Z,U,E,L1,C1,V2,BP,BA,CM)   |
|         |                         ✕(S)                        |               ✕()                |
+---------+-----------------------------------------------------+----------------------------------+
+---------+----------------------+----------------------+
|         |        ngtcp2        |       openssl        |
+---------+----------------------+----------------------+
|  ngtcp2 | G: 9059 (± 74) kbps  | G: 9234 (± 14) kbps  |
|         | C: 5067 (± 211) kbps | C: 3675 (± 195) kbps |
+---------+----------------------+----------------------+
| openssl | G: 9109 (± 90) kbps  |  G: 9362 (± 0) kbps  |
|         | C: 4572 (± 69) kbps  | C: 2322 (± 103) kbps |
+---------+----------------------+----------------------+

```

I'll investigate the failing case shortly, but I think this looks pretty good now!

nhorman:
the remaining failing test case is in the quic-interop retry test, and appears to be due to ngtcp2 sending multiple retry frames in response to client hellos that span multiple datagrams.  I've opened a separate issue to track it here:
https://github.com/ngtcp2/ngtcp2/issues/1585

tatsuhiro-t:
To avoid any possible confusion, I close this PR.  Adding openssl backend support, if ever happens, will be done via #1582

PR #1600, opened 2025-04-10
Skip packet number

tatsuhiro-t:


redbaron:
Just for my education, what is motivation for packet number skip?

PR #1608, opened 2025-04-11
CMake: Add ENABLE_LIB_ONLY option

frankosterfeld:
Disables examples and thus avoids external dependencies.


tatsuhiro-t:

PR #1666, opened 2025-06-06
examples: Refactor siphash key generation

### LINKED ISSUE #1665: hash quality for ngtcp2_cid in example code seems to depend on whether NDEBUG is set
Dear ngtcp2 devs,

The `std::hash` specialization for `ngtcp2_cid` in the example code relies on a keyed hash function.
If assertions are disabled, and assuming I have a correct understanding of how `assert` works, the statement inside the assert is never be executed, so the hash key will contain uninitialized memory in that case. Hence, I'd say this gives a wrong example to users of the library.

Existing code:
```cpp
namespace std {
template <> struct hash<ngtcp2_cid> {
  hash() {
    assert(0 == ngtcp2::util::generate_secure_random(
                  as_writable_uint8_span(std::span{key})));
  }

  std::size_t operator()(const ngtcp2_cid &cid) const noexcept {
    return static_cast<size_t>(siphash24(key, {cid.data, cid.datalen}));
  }

  std::array<uint64_t, 2> key;
};
} // namespace std
```

I would suggest:
```cpp
namespace std {
template <> struct hash<ngtcp2_cid> {
  hash() {
    int ret = ngtcp2::util::generate_secure_random(
                  as_writable_uint8_span(std::span{key}));
    assert(0 == ret);
  }

  std::size_t operator()(const ngtcp2_cid &cid) const noexcept {
    return static_cast<size_t>(siphash24(key, {cid.data, cid.datalen}));
  }

  std::array<uint64_t, 2> key {}; // zero-initialize here by means of {} ?
};
} // namespace std
```

tatsuhiro-t:
Fixes #1665

### DIFF
--- a/examples/util.cc
+++ b/examples/util.cc
@@ -833,6 +833,25 @@ read_ech_server_config(const std::string_view &path) {
     .ech_config = std::move(*ech_config),
   };
 }
+
+std::span<uint64_t, 2> generate_siphash_key() {
+  static auto key = []() {
+    std::array<uint64_t, 2> key;
+
+    auto rv = generate_secure_random(as_writable_uint8_span(std::span{key}));
+    if (rv != 0) {
+      assert(0);
+      abort();
+    }
+
+    return key;
+  }();
+
+  ++key[0];
+
+  return key;
+}
+
 } // namespace util
 
 std::ostream &operator<<(std::ostream &os, const ngtcp2_cid &cid) {
--- a/examples/util.h
+++ b/examples/util.h
@@ -552,6 +552,8 @@ struct ECHServerConfig {
 std::optional<ECHServerConfig>
 read_ech_server_config(const std::string_view &path);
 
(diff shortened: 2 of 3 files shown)


examples/util.cc
```diff
@@ -833,6 +833,25 @@ read_ech_server_config(const std::string_view &path) {
...
+    auto rv = generate_secure_random(as_writable_uint8_span(std::span{key}));
+    if (rv != 0) {
+      assert(0);
+      abort();
+    }
+
+    return key;
+  }();
+
+  ++key[0];
```
niekbouman-rl, line 850: @tatsuhiro-t 
out of curiosity, why is this line needed?

tatsuhiro-t: To get new key for each map.

niekbouman-rl: Okay, I see, thanks.

PR #1689, opened 2025-06-30
Fix pc files for static  linkage

talregev:
Fix pc files for static  linkage.
I try this with ngtcp2 + openssl + nghttp3 + curl to compile curl with http3. on linux static compilation (with vcpkg).
I verify that in vcpkg:
openssl, libressl, boringssl create pc file name: openssl.pc
wolfssl -> wolfssl.pc
libgnutls -> gnutls.pc 

Without this fix I get error compilation.
ngtcp2 + openssl + nghttp3 + curl to compile curl with http3.
I get these errors:

```
/usr/bin/ld: /home/tal/vcpkg/installed/x64-linux/debug/lib/libngtcp2_crypto_ossl.a(ossl.c.o): in function `ngtcp2_crypto_set_local_transport_params':
/home/tal/vcpkg/buildtrees/ngtcp2/src/v1.13.0-ba2cdf4128.clean/crypto/ossl/ossl.c:943: undefined reference to `SSL_set_quic_tls_transport_params'
/usr/bin/ld: /home/tal/vcpkg/installed/x64-linux/debug/lib/libngtcp2_crypto_ossl.a(ossl.c.o): in function `crypto_ossl_configure_session':
/home/tal/vcpkg/buildtrees/ngtcp2/src/v1.13.0-ba2cdf4128.clean/crypto/ossl/ossl.c:1178: undefined reference to `SSL_set_quic_tls_cbs'
```

### DIFF
--- a/crypto/boringssl/libngtcp2_crypto_boringssl.pc.in
+++ b/crypto/boringssl/libngtcp2_crypto_boringssl.pc.in
@@ -31,3 +31,4 @@ URL: https://github.com/ngtcp2/ngtcp2
 Version: @VERSION@
 Libs: -L${libdir} -lngtcp2_crypto_boringssl
 Cflags: -I${includedir}
+Requires: openssl
--- a/crypto/gnutls/libngtcp2_crypto_gnutls.pc.in
+++ b/crypto/gnutls/libngtcp2_crypto_gnutls.pc.in
@@ -31,3 +31,4 @@ URL: https://github.com/ngtcp2/ngtcp2
 Version: @VERSION@
 Libs: -L${libdir} -lngtcp2_crypto_gnutls
 Cflags: -I${includedir}
+Requires: gnutls
--- a/crypto/ossl/libngtcp2_crypto_ossl.pc.in
+++ b/crypto/ossl/libngtcp2_crypto_ossl.pc.in
@@ -31,3 +31,4 @@ URL: https://github.com/ngtcp2/ngtcp2
 Version: @VERSION@
 Libs: -L${libdir} -lngtcp2_crypto_ossl
 Cflags: -I${includedir}
+Requires: openssl
--- a/crypto/quictls/libngtcp2_crypto_quictls.pc.in
+++ b/crypto/quictls/libngtcp2_crypto_quictls.pc.in
@@ -31,3 +31,4 @@ URL: https://github.com/ngtcp2/ngtcp2
 Version: @VERSION@
 Libs: -L${libdir} -lngtcp2_crypto_quictls
 Cflags: -I${includedir}
+Requires: openssl
--- a/crypto/wolfssl/libngtcp2_crypto_wolfssl.pc.in
+++ b/crypto/wolfssl/libngtcp2_crypto_wolfssl.pc.in
@@ -31,3 +31,4 @@ URL: https://github.com/ngtcp2/ngtcp2
 Version: @VERSION@
 Libs: -L${libdir} -lngtcp2_crypto_wolfssl
 Cflags: -I${includedir}
+Requires: wolfssl


vszakats:
It looks like another "picky linker" case for binutils `ld`, where it's unable
to resolve symbols without passing them in reverse dependency order,
or listing them twice (or passing them between `-Wl,--start-group` and
`-Wl,--end-group`).

The patch looks correct to me to address this.

Ref: https://github.com/curl/curl/discussions/17600

tatsuhiro-t:

vszakats:
This change turned out breaking the GnuTLS HTTP/3 CI job in curl:
```
configure: error: --with-ngtcp2 was specified but could not find ngtcp2_crypto_gnutls pkg-config file.
```
https://github.com/curl/curl/actions/runs/16525946073/job/46744939388?pr=18022#step:14:273

Tried `Require.private:` but it did not fix that build. Deleting
the `Require:` line did. But I'm unsure what would be the correct
fix, and why `pkg-config --exists gnutls` fails, while the package
is listed and detected according to debug output.

https://github.com/curl/curl/pull/18028

talregev:
Can you check when you compile gnutls, you get the pc file `gnutls.pc`?
Can you check that `gnutls.pc` is on the `PKG_CONFIG_PATH`?

talregev:
> pkg-config --exists gnutls fails

Maybe it mean that `gnutls.pc` file is not on the `PKG_CONFIG_PATH`

vszakats:
It's there as confirmed via ls, the content of PKG_CONFIG_PATH, pkg-config --debug, --list-all.

talregev:
Can you past the content of the `gnutls.pc` file?
Maybe it missing it's required.

tatsuhiro-t:
This is not a fix for the issue, but I think we should use Requires.private instead of Requires because the dynamic linking does not need -lgnutls.  We only need it on static linking, which is achieved by Requires.private.

vszakats:
@talregev Thanks for the pointers!

After tracing the curl configure script, it now seems this is a curl issue,
combined with `pkg-config` behavior. The configure scripts sets
`PKG_CONFIG_LIBDIR` right before calling `--exists`, which _overrides_
any previously set `PKG_CONFIG_DIR` when using `pkg-config` (v1.8.1).
`pkgconf` would likely work as expected, because `PKG_CONFIG_LIBDIR`
_extends_ (not overrides) the `PKG_CONFIG_DIR` there.

https://github.com/curl/curl/blob/cb9b1a4c4e875ac4eac2209d8686acef3114abdf/acinclude.m4#L1376-L1381
https://manpages.debian.org/unstable/pkg-config/pkg-config.1.en.html
https://manpages.debian.org/unstable/pkgconf/pkgconf.1.en.html

If my assessment is correct, I'm surprised why wasn't it causing issues in
the past. Perhaps it's the first time a pkg-config module (libngtcp2_crypto_gnutls)
with custom path is depending on modules (gnutls and its deps) residing
on the global pkg-config path!

talregev:
I am happy that you solve the problem, and I hope to see the fix! 💪🏻

talregev:
> This is not a fix for the issue, but I think we should use Requires.private instead of Requires because the dynamic linking does not need -lgnutls. We only need it on static linking, which is achieved by Requires.private.

I agree. Thank you for the PR.

vszakats:
9 months later I'm still debugging this!

For GnuTLS it's a curl autotools build issue. Haven't yet pulled the trigger
on the [patch](https://github.com/curl/curl/pull/20920), as it requires an exception compared to other dependencies.

For BoringSSL, it looks like an ngtcp2 / BoringSSL issue, depending on how we
look at it. After this patch `ngtcp2_crypto_boringssl.pc` depends on `openssl.pc`,
but BoringSSL does NOT ship an `openssl.pc`. It means when everything is 
correctly setup, it fails detection because of this missing dependency. To make
it work, pkg-config needs to find a fake/surrogate `openssl.pc` to pass the check,
then the build proceed use the correct BoringSSL libs regardless. This is probably
easy to overlook because most envs have a (non-BoringSSL) `openssl.pc` laying
around at a system location.

For more empuzzlement, this patch _is_ accurate when using AWS-LC, which
does provide its `openssl.pc`:
https://github.com/aws/aws-lc/blob/f3c6fff9a1a04183b5fbacdf55a2abf9a29c142f/CMakeLists.txt#L1545-L1550

I don't anticipate BoringSSL adding `openssl.pc` anytime soon, so one way
to fix this would be to split off AWS-LC into its own crypto module, while keeping
everything as-is, except the name (similarly to the quictls/LibreSSL split
earlier), then revert this patch for `ngtcp2_crypto_boringssl.pc.in`. This has
the additional benefit to support BoringSSL and AWS-LC crypto libs in parallel,
and may remove confusion once the libs diverge enough to require different
codepaths (last time I checked this wasn't yet the case).

https://github.com/curl/curl/pull/20926#issuecomment-4063674714

talregev:
> > Wow, sound harsh. I happy you find a some workaround. I understand how you feel about it. Are you going to do PR in ngtcp2?
> 
> If this sounds acceptable, I might, yes. For now it's quite fresh and I need to let it settle a bit, double-check, etc (e.g. ATM I don't understand why AWS-LC works in practice in curl, because it also shouldn't, due to configure script reasons.) I'm hoping it will happen while digging myself out of the stack of PRs dealing with this.

I am sending you energies to understand it better. I will also be happy to see your PRs solution soon.

PR #1716, opened 2025-07-20
build: use distinct crypto library name for LibreSSL

vszakats:
To allow building for quictls and LibreSSL under the same prefix
without conflict. This helps packaging by allowing these to crypto libs
to coexist.

After this patch, `ngtcp2_crypto_libressl` lib (and matching `.pc` file)
is built when detecting LibreSSL, instead of the previous
`ngtcp2_crypto_quictls`, which was shared with quictls builds.

Notice that this introduces an incompatibility at the build level, and
will require changes when consuming ngtcp2 when built against LibreSSL.

- Before this patch:
  both quictls and LibreSSL-built crypto libs and pkg-config files were
  named:
  `libngtcp2_crypto_quictls.*`

- After this patch:
  quictls-built crypto lib keeps the existing name:
  `libngtcp2_crypto_quictls.*`.

  LibreSSL-built crypto libs gets its own distinct name:
  `libngtcp2_crypto_libressl.*`.


tatsuhiro-t:
This surely breaks the backward compatibility where previously user expects that libngtcp2_crypto_quictls is linked to libressl.  What is the reasoning of this patch and a use case?

vszakats:
The use case is to seamlessly use the same ngtcp2 package to build
against both quictls and libressl.

Currently, packages need to chose between one of them, and the other
has to be built manually and then configured differently because it must
reside under a different prefix. Also, at the moment it's not possible for 
curl's FindNGTCP2 module to automatically select the correct crypto
library via `find_package(NGTCP2 REQUIRED "quictls")` because this
might point to either LibreSSL or quictls depending on how ngtcp was
built.

Perhaps it's not too late to do this, and it'd make life, and LibreSSL
support easier down the line, even with the compatibility break, I think.

tatsuhiro-t:
This is the unfortunate situation of openssl forks.  They basically replace each other.
Boringssl has been treated differently because it provided static libraries when it was first introduced.
Currently we assume that user should choose quictls or libressl.
libressl is quite limited in its capability, which I think is intentional, only used for specific situations like Apple products.  I think it is quite rare to build both quictls and libressl at the same time.  Even with this patch, autotools build cannot build them at the same time.

vszakats:
> This is the unfortunate situation of openssl forks. They basically replace each other. Boringssl has been treated differently because it provided static libraries when it was first introduced. Currently we assume that user should choose quictls or libressl. libressl is quite limited in its capability, which I think is intentional, only used for specific situations like Apple products. I think it is quite rare to build both quictls and libressl at the same time. Even with this patch, autotools build cannot build them at the same time.

Building them at the same is a different issue. It'd be nice to
do it that way, but so far I'm resorting to building separately 
and installing to the same prefix. This works fine. Building it in
one go probably would be possible but would need quite a bit
of build gymnastics, and may make the configurationunusual
so I didn't go for it. The architecture would allow it though.

curl's official Windows binaries use LibreSSL. We also threat 
LibreSSL on the same level as other backends in CI. It's quite
important. It's an impossible choice to decide between quictls
and LibreSSL, each having their own pros and cons.

vszakats:
> But you chose libressl over quictls for curl windows build, right? Why do you still need them both?

I don't need quictls for the official Windows builds now.

But, I do a lot of work around curl's builds and build combinations,
where quictls and libressl are both built and used extensively.

It helps a lot to switch between them seamlessly in H3 ngtcp2 builds.

On the linked Homebrew thread adding the ngtcp2 package, we're
conversing about possibly supporting not just `ossl` but maybe
also the other backends that Homebrew supports. This is probably
doable, except for quictls and LibreSSL, unless Homebrew makes
a hard choice in favor of one of them. Otherwise they can only be
supported in conflicting packages. (Homebrew already offers both
quictls and libressl as packages)

Conflicting packages are a nuisance because one needs to uninstall
or 'unlink' one package before installing/'linking' the other.

We've also been playing with build-fuzzers in curl, which configures
and builds random configurations, looking for fallouts. In such script
it helps a great deal if all dependencies are available as-is.

vszakats:
> Thanks. Installing multiple openssl forks at the same time might cause serious trouble that are somewhat hard to track, but that is the decision of homebrew.
> 
> We actively say backward compatibility is important for build tools. We say that for all PRs where cmake is involved. Now we are actively breaking it for optional features. In other words, if we have a good reasoning behind this (and it should be written in the commit message), I think it would be fine.

Thanks! I'll try extending the text with more information
about the motivation, and highlight the incompatible nature.

> I am also concerned about the future of quictls. It has been very nice to provide QUIC API for years. But now the folks behind it stop maintaining initial forks (the final version is v3.1.7+quic, and v1.1.1w which is the best openssl ever made, both are outdated by now). And they started creating permanent fork and there are no release yet. I think its value is diminishing quite rapidly now that openssl 3.5 finally added QUIC API which is needlessly hard to use though.

I agree. An earlier lapse similar to this was one of the triggers for
switching to LibreSSL in curl-for-win. As you say, with OpenSSL 3.5,
the reasons to use quictls are slimmer than before. Its original
QUIC API has a few upsides compared to the OpenSSL iteration
(AFAIK), but that's unlikely to change the general direction.

configure.ac
```diff
@@ -512,7 +540,8 @@ if test "x${lib_only}" = "xno" &&
 fi
 
 AM_CONDITIONAL([ENABLE_EXAMPLES], [ test "x${enable_examples}" = "xyes" ])
-AM_CONDITIONAL([ENABLE_EXAMPLE_QUICTLS], [ test "x${enable_examples}" = "xyes" && test "x${have_quictls}" = "xyes" ])
+AM_CONDITIONAL([ENABLE_EXAMPLE_QUICTLS], [ test "x${enable_examples}" = "xyes" && test "x${have_quictls}" = "xyes" -o "x${have_libressl}" = "xyes" ])
```
tatsuhiro-t: I think `-o "x${have_libressl}" = "xyes"` is not required because libressl has its own ENABLE_EXAMPLE_LIBRESSL below.

vszakats: It has, but the logic building the examples is still guarded by `ENABLE_EXAMPLE_QUICTLS`, so it also needs to be enabled. Maybe there is a cleaner way to do this, but I didn't want to clone the whole section with almost identical content.

tatsuhiro-t: Just adding `|| ENABLE_EXAMPLE_LIBRESSL` to `if ENABLE_EXAMPLE_QUICTLS` in examples/Makefile.am is enough, I guess?

vszakats: AFAIK automake's `if` statement doesn't support complex expressions :( (I wish this was possible, it'd make life so much easier.)

https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html

vszakats:
Updated the PR text with a bit more detail. I'm also open to suggestions for improvements.

configure.ac
tatsuhiro-t: Thanks.  Still you should do at least something like so:

```suggestion
AM_CONDITIONAL([ENABLE_EXAMPLE_QUICTLS], [ test "x${enable_examples}" = "xyes" && (test "x${have_quictls}" = "xyes" || test "x${have_libressl}" = "xyes") ])
```

vszakats: Updated and pushed!

tatsuhiro-t (approved):

PR #1752, opened 2025-08-23
build: install crypto header for libressl

vszakats:
Follow-up to 433d297aacb93abf0ccddcd678d09239dfe4dd1e #1716

### DIFF
--- a/crypto/includes/CMakeLists.txt
+++ b/crypto/includes/CMakeLists.txt
@@ -25,7 +25,7 @@ install(FILES
     ngtcp2/ngtcp2_crypto.h
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/ngtcp2")
 
-if(HAVE_QUICTLS)
+if(HAVE_QUICTLS OR HAVE_LIBRESSL)
   install(FILES
       ngtcp2/ngtcp2_crypto_quictls.h
       DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/ngtcp2")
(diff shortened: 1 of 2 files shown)


tatsuhiro-t:

vszakats:
Thank you @tatsuhiro-t!

tatsuhiro-t:
If you make a fresh install ngtcp2 with only libressl crypto enabled, then I think yes, the header file is missing.

botovq:
Ok, thank you. I need to figure out how to deal with it then... We need to adapt ~70 OpenBSD ports to use the new library name and it would be nice to do that with some time to spare before the release freeze in ~1 month.

vszakats:
> Ok, thank you. I need to figure out how to deal with it then... We need to adapt ~70 OpenBSD ports to use the new library name and it would be nice to do that with some time to spare before the release freeze in ~1 month.
> 
> (It's not a big deal, I just need to think a little bit. It's much appreciated that this library supports libressl natively.)

This was my temporary workaround in curl-for-win, if that's of any help:
```diff
--- a/ngtcp2.sh
+++ b/ngtcp2.sh
@@ -58,6 +58,9 @@ _VER="$1"
   cmake --build "${_BLDDIR}"
   cmake --install "${_BLDDIR}" --prefix "${_PPS}"
 
+  # PR: https://github.com/ngtcp2/ngtcp2/pull/1752
+  [ "${_VER}" = '1.15.0' ] && cp -p crypto/includes/ngtcp2/ngtcp2_crypto_quictls.h "${_PPS}"/include/ngtcp2/
+
   # Delete .pc files
   rm -r -f "${_PPS}"/lib/pkgconfig
```
https://github.com/curl/curl-for-win/commit/3656a860bd9ca4d4e2693100b9d51546bcfae38e

I apologize for overlooking this build-case in the original patch.

PR #1779, opened 2025-09-10
cmake: do not require C++ for lib-only non-BoringSSL builds

vszakats:
C++ is only used to build examples and for a feature check when building
against BoringSSL.


CMakeLists.txt
```diff
@@ -114,7 +118,12 @@ if(NOT CMAKE_C_COMPILER_ID MATCHES "MSVC")
     cmake_push_check_state()
     set(CMAKE_REQUIRED_LIBRARIES "-fsanitize=address")
     check_c_compiler_flag(-fsanitize=address C__fsanitize_address_VALID)
-    check_cxx_compiler_flag(-fsanitize=address CXX__fsanitize_address_VALID)
+    if(NOT ENABLE_LIB_ONLY)
```
tatsuhiro-t: This should be `if(ENABLE_LIB_ONLY)` because the comment says "Bypass ..." in this condition.

vszakats: Of course, nice catch!

PR #1789, opened 2025-09-17
configure.ac: allow picotls without ossl

CL-Jeremy:
This PR decouples detection of flavored OpenSSL from that of vanilla OpenSSL. I suppose this behavior is in CMake already.

Before this PR, it is impossible not to also build `libngtcp2_crypto_ossl` while building for Picotls. With this PR, the following is allowed:

```
./configure --with-openssl=no --with-picotls=yes
```

The effective change is minimal, as viewable with `git diff -b`. Only the change in indentation made the diff much larger.

I'm fairly certain that other backends are not affected by this, though not 100%. Advices are much welcome.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -210,7 +210,8 @@ have_vanilla_openssl=no
 have_quictls=no
 have_libressl=no
 have_ossl=no
-if test "x${request_openssl}" != "xno"; then
+if test "x${request_openssl}" != "xno" ||
+   test "x${request_picotls}" = "xyes"; then
   PKG_CHECK_MODULES([OPENSSL], [openssl >= 1.1.1],
                     [have_openssl=yes], [have_openssl=no])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -210,7 +210,8 @@ have_vanilla_openssl=no
 have_quictls=no
 have_libressl=no
 have_ossl=no
-if test "x${request_openssl}" != "xno"; then
+if test "x${request_openssl}" != "xno" ||
+   test "x${request_picotls}" = "xyes"; then
```
tatsuhiro-t, line 214: This might be confusing at first glance that openssl detection kicks in even if openssl=no.
But this is some kind of indirect dependency from picotls, and we only support picotls with openssl.
I think this is fine for now.

tatsuhiro-t:

PR #1811, opened 2025-10-02
Replace POPCNT-based ispow2 with portable bitwise implementation to prevent illegal instruction crashes

### LINKED ISSUE #1810: Crash on CPUs without POPCNT due to unconditional popcount use in ispow2 (asserts forced on in all builds)
### Summary
Description: The application crashes with an illegal instruction (POPCNT) on older or virtualized x86/x64 CPUs that do not support the POPCNT instruction (CPUID ECX bit 23 = 0).
The failure originates in the debug helper ispow2, which used POPCNT intrinsics inside #ifndef NDEBUG. However, the build system (CMakeLists.txt) explicitly strips -DNDEBUG from Release / MinSizeRel / RelWithDebInfo configurations:
```CMake
# Do not disable assertions based on CMAKE_BUILD_TYPE.
foreach(_build_type "Release" "MinSizeRel" "RelWithDebInfo")
  foreach(_lang C CXX)
    string(TOUPPER "CMAKE_${_lang}_FLAGS_${_build_type}" _var)
    string(REGEX REPLACE "(^| )[/-]D *NDEBUG($| )" " " ${_var} "${${_var}}")
  endforeach()
endforeach()
```
As a result, the “debug-only” POPCNT code ran in production builds. On CPUs lacking POPCNT support, execution of the intrinsic caused an immediate crash during assertion checks (e.g. validating ring buffer sizes).

#### Impact:
•	Release builds can crash at startup or during buffer operations on non-POPCNT hardware.
•	Hard to diagnose without symbols; happens before higher-level logic runs.

#### Expected Behavior:
A power-of-two assertion should not execute unsupported instructions and should work on all x86/x64 CPUs.
#### Actual Behavior:
Unconditional POPCNT intrinsic leads to illegal instruction fault.

mk185147:
Resolves #1810. This PR removes all POPCNT intrinsic usage from the debug-only ispow2 helper in ngtcp2_ringbuf.c, replacing it with the canonical bitwise power-of-two test: n && !(n & (n - 1)). This resolves crashes on CPUs without POPCNT while preserving performance.
#### Root Cause:
•	ispow2 used POPCNT intrinsics under #ifndef NDEBUG.
•	CMakeLists.txt strips -DNDEBUG from Release-like builds, so the code executed everywhere.
•	On CPUs lacking POPCNT, the intrinsic produced an illegal instruction fault.
#### Why This Implementation:
•	The bitwise formulation is architecture-neutral, constant-time, and emits just a few basic instructions.
•	For a single power-of-two check, POPCNT provides negligible benefit versus its complexity.
•	Avoids runtime CPUID dispatch and platform-specific branches.
•	Eliminates risk of crashes on older or constrained environments (legacy hardware, minimal VMs).
#### Benchmarks (Windows, C++20 test harness):
64-bit (ns/element):
•	Bitwise: 1.58346
•	__popcnt64: 1.52366
•	_mm_popcnt_u64: 1.50899
•	std::popcount: 1.78225
32-bit (ns/element):
•	Bitwise: 9.47397
•	__popcnt: 9.26461
•	std::popcount: 10.3988
Differences are marginal; bitwise is close to hardware POPCNT and faster than fallback or std::popcount cases.
#### Changes Made:
•	Removed previous POPCNT and runtime dispatch logic.
•	Added a single debug-only implementation:
```
#ifndef NDEBUG
    static int ispow2(size_t n) { return n && !(n & (n - 1)); }
#endif
```
•	No functional changes elsewhere.
#### Safety / Risk:
Low. Used only in assertions; logic is well-known and widely used. No ABI changes.
#### Performance Impact:
Neutral to positive. Simplifies code and avoids feature detection overhead.
#### Future Considerations:
If a future hot path in release code requires ispow2, reuse the same bitwise form; add POPCNT only if profiling proves a measurable win across supported CPUs.

### DIFF
--- a/lib/ngtcp2_ringbuf.c
+++ b/lib/ngtcp2_ringbuf.c
@@ -32,18 +32,41 @@
 #include "ngtcp2_macro.h"
 
 #ifndef NDEBUG
-static int ispow2(size_t n) {
+/* Provide fastest path when POPCNT is available: we detect once and then
+   call the selected implementation without an extra branch each time. */
 #  if defined(_MSC_VER) && !defined(__clang__) &&                              \
     (defined(_M_ARM) || (defined(_M_ARM64) && _MSC_VER < 1941))
+static int ispow2(size_t n) { /* Simple portable fallback */
   return n && !(n & (n - 1));
-#  elif defined(WIN32)
-  return 1 == __popcnt((unsigned int)n);
-#  else  /* !((defined(_MSC_VER) && !defined(__clang__) && (defined(_M_ARM) || \
-            (defined(_M_ARM64) && _MSC_VER < 1941))) || defined(WIN32)) */
-  return 1 == __builtin_popcount((unsigned int)n);
-#  endif /* !((defined(_MSC_VER) && !defined(__clang__) && (defined(_M_ARM) || \
-            (defined(_M_ARM64) && _MSC_VER < 1941))) || defined(WIN32)) */
 }
+#  elif defined(WIN32)
+#    if defined(_M_IX86) || defined(_M_X64)
+static int ispow2_popcnt(size_t n) { return 1 == __popcnt((unsigned int)n); }
+#    endif /* x86/x64 */
+static int ispow2_fallback(size_t n) { return n && !(n & (n - 1)); }
+#    if defined(_M_IX86) || defined(_M_X64)
+static int ispow2_runtime(size_t n) {
+  int info[4] = {0};
+  __cpuid(info, 1);
+  /* ECX bit 23 indicates POPCNT support */
+  if (info[2] & (1 << 23)) {
+    /* Publish chosen implementation; benign data race acceptable in debug */
+    extern int (*ngtcp2_ispow2_impl)(size_t); /* forward */
+    ngtcp2_ispow2_impl = ispow2_popcnt;
+  } else {
+    extern int (*ngtcp2_ispow2_impl)(size_t);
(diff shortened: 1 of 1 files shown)


lib/ngtcp2_ringbuf.c
```diff
@@ -32,19 +32,9 @@
...
-#  else  /* !((defined(_MSC_VER) && !defined(__clang__) && (defined(_M_ARM) || \
-            (defined(_M_ARM64) && _MSC_VER < 1941))) || defined(WIN32)) */
-  return 1 == __builtin_popcount((unsigned int)n);
-#  endif /* !((defined(_MSC_VER) && !defined(__clang__) && (defined(_M_ARM) || \
-            (defined(_M_ARM64) && _MSC_VER < 1941))) || defined(WIN32)) */
-}
-#endif /* !defined(NDEBUG) */
+/* Power-of-two test; simple portable bit trick. */
+static int ispow2(size_t n) { return n && !(n & (n - 1)); }
+#endif /* !NDEBUG */
```
tatsuhiro-t:
```suggestion
#endif /* !defined(NDEBUG) */
```

This is what we do consistently in this codebase as the original code did.

tatsuhiro-t (changes requested):

lib/ngtcp2_ringbuf.c
mk185147: Fixed by reverting back

PR #1877, opened 2025-11-02
Delete unusable initial rx key on if tx key installation fails

tatsuhiro-t:


tatsuhiro-t:
Note that there are no memory leaks in these functions in any execution paths.

PR #1910, opened 2025-11-18
 Fix: Resolve linker failure in examplestest by conditionally compiling simulation tests

devkdas:
The `examplestest` executable unconditionally includes and references the `ngtcp2::sim_suite`, which has a hard dependency on wolfSSL.

This causes a linker error (`undefined reference`) when the project is configured and built without wolfSSL support, as the corresponding source files for the simulation test are not compiled.

This patch wraps the relevant code in `examplestest.cc` with an `#if ENABLE_EXAMPLE_WOLFSSL` directive. This ensures that the simulation test suite is only compiled and linked when wolfSSL is explicitly enabled, resolving the build failure.

### DIFF
--- a/examples/examplestest.cc
+++ b/examples/examplestest.cc
@@ -38,12 +38,12 @@
 
 int main(int argc, char *argv[]) {
   const MunitSuite suites[] = {
-      ngtcp2::util_suite,
-      ngtcp2::siphash_suite,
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  It seems like you use wrong macros that are never defined.
Superseded by #1911

Issue #1961, opened 2025-12-07
No exported CMake targets for crypto libs


dg0yt:
For the crypto libs, there are pkg-config file, but similar imported targets are missing from the exported CMake config.

Noticed while trying to use ngtcp2 CMake config with the curl CMake build and static library linkage (vcpkg x64-linux). I don't want the CMake build to import OpenSSL once via CMake and once via pkg-config.

dg0yt:
How should such targets be named?
With the existing export set, there would be e.g. `ngtcp2::ngtcp2_crypto_ossl_static`.

tatsuhiro-t:
I think it would be nice to name it like we already do in cmake.  I mean similar name convention is good.
I should also mention that keeping backward compatibility is a must.

talregev:
@dg0yt Can you elaborate why the suffix `_static` is needed?

talregev:
I vote for this option:
`ngtcp2::crypto_ossl_static (modifying EXPORT_NAME property)`

tatsuhiro-t:
@vszakats Do you have any concern about introducing namespace?

dg0yt:
> [@dg0yt](https://github.com/dg0yt) Can you elaborate why the suffix `_static` is needed?

This project allows to build static and shared libs at the same time, and this is how they do it.

(I consider this a CMake antipattern, but I respect the project's choices.)

dg0yt:
I really didn't expect the namespace to be a topic of discussion.
- The namespace doesn't take away anything which exists now.
- The namespace isn't new, but already used for the exported ngtcp2 target.
- Namespaces in CMake improve diagnostics for targets from dependencies. (CMake can recognize that an identifier should be a CMake target with properties, not a bareword to be passed to the linker.)

Please focus on the user-facing name after the `::`.

talregev:
this already solved?

dg0yt:
https://github.com/ngtcp2/ngtcp2/commits/main/crypto
The OpenSSL case is covered since https://github.com/ngtcp2/ngtcp2/pull/1963.
Other backends still need to be done.

dg0yt:
Not stale.

dg0yt:
Not stale, but I cannot reopen.

PR #1963, opened 2025-12-07
Export CMake target for ngtcp2_crypto_ossl(_static)

dg0yt:
POC for #1961, openssl-only at the moment -> Draft.

For simplicity, this change reuses the existing `${PROJECT_NAME}Targets` export set.
- It might make sense to move the `install(EXPORT ...)` and the corresponding `config.cmake.in` from `lib` to the top-level directory.
- Transitive dependencies like OpenSSL are subject to the actual crypto backend used in the downstream project. It might make sense to offer a `find_package(... COMPONENTS ...)` interface to let the consumers filter backends and dependencies.

By switching to `PROJECT_SOURCE_DIR`/`PROJECT_BINARY_DIR` for include path construction, the directories get rid of the `../../` components.

### DIFF
--- a/crypto/ossl/CMakeLists.txt
+++ b/crypto/ossl/CMakeLists.txt
@@ -29,11 +29,11 @@ set(ngtcp2_crypto_ossl_SOURCES
 )
 
 set(ngtcp2_crypto_ossl_INCLUDE_DIRS
-  "${CMAKE_CURRENT_SOURCE_DIR}/../../lib"
-  "${CMAKE_CURRENT_SOURCE_DIR}/../../crypto/includes"
-  "${CMAKE_CURRENT_BINARY_DIR}/../../crypto/includes"
-  "${CMAKE_CURRENT_SOURCE_DIR}/../../crypto"
-  "${CMAKE_CURRENT_BINARY_DIR}/../../crypto"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/lib>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/crypto/includes>"
+  "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/crypto/includes>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/crypto>"
+  "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/crypto>"
 )
 
 foreach(name libngtcp2_crypto_ossl.pc)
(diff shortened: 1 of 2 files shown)


crypto/ossl/CMakeLists.txt
```diff
@@ -29,14 +29,11 @@ set(ngtcp2_crypto_ossl_SOURCES
...
-  "${CMAKE_CURRENT_BINARY_DIR}/../../crypto/includes"
-  "${CMAKE_CURRENT_SOURCE_DIR}/../../crypto"
-  "${CMAKE_CURRENT_BINARY_DIR}/../../crypto"
-  "${OPENSSL_INCLUDE_DIRS}"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/lib>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/crypto/includes>"
+  "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/crypto/includes>"
+  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/crypto>"
+  "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/crypto>"
 )
```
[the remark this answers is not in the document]
dg0yt: No.

crypto/ossl/CMakeLists.txt
```diff
@@ -55,9 +52,10 @@ if(ENABLE_SHARED_LIB)
   )
   target_include_directories(ngtcp2_crypto_ossl PUBLIC
     ${ngtcp2_crypto_ossl_INCLUDE_DIRS})
-  target_link_libraries(ngtcp2_crypto_ossl ngtcp2 ${OPENSSL_LIBRARIES})
+  target_link_libraries(ngtcp2_crypto_ossl PUBLIC ngtcp2 OpenSSL::SSL)
```
dg0yt, line 55: `ngtcp2_crypto_ossl` links `ngtcp2`. (Unchanged.) But `libngtcp2_crypto_ossl.pc` does not pull in ngtcp2 in `Libs(.private)` or `Requires(.private)`. Is there an actual binary dependency?

tatsuhiro-t: Yes.  I think we forgot to add libngtcp2 to Requires.private for libngtcp2_crypto.  Will fix it.

tatsuhiro-t: #1964

dg0yt:
FTR this is part of WIP preparation for curl 8.18 in vcpkg. 
vcpkg simply needs first-class CMake config where possible.

tatsuhiro-t:
@vszakats What do you think about this change?

PR #1979, opened 2026-01-06
examples: Avoid strict aliasing violation

tatsuhiro-t:


examples/client.cc
```diff
@@ -1049,16 +1049,17 @@ ngtcp2_ssize write_pkt(ngtcp2_conn *conn, ngtcp2_path *path,
 ngtcp2_ssize Client::write_pkt(ngtcp2_path *path, ngtcp2_pkt_info *pi,
                                uint8_t *dest, size_t destlen,
                                ngtcp2_tstamp ts) {
-  std::array<nghttp3_vec, 16> vec;
+  std::array<SharedVec, 16> vec;
```
thesamesam, line 1052: For unions, you can't pun because of the  'active member' in strict C++ but it's an extension in GCC and Clang though. That said, not sure it matters much because MSVC doesn't do TBAA.

But in any case, you have to always take the access via the union member, i.e. you need to be be doing `element.v2` and `element.v3`, not `reinterpret_cast` to one of the types within the union AFAIK.

And that would let you drop `reinterpret_cast` (usually you want one of a union _or_ `reinterpret_cast`, not both).

jwakely: Assuming that `nghttp3_conn_writev_stream` looks at more than one element of the array, using a union doesn't fix the code.

Pointer arithmetic on `nghttp3_vec*` is undefined if it doesn't point to a real array of type `nghttp3_vec[N]`. An array of `SharedVec` is not an array of `nghttp3_vec`. https://eel.is/c++draft/expr.add#6

tatsuhiro-t: Thank you.  Will rever this PR.  Can C++23 std::start_lifetime_as_array help here?

PR #2190, opened 2026-05-19
Fix incorrect initialization of remote unidirectional streams when they are opened by a reset-stream message

zalan24:
I have found a bug in my integration tests, where some streams would never get released. This turned out to be due to bad initialization in a code path, where the remote unidirectional stream's writing end is not shut down as it is supposed to be.

The streams are opened on different code paths depending on what kind of message they receive first. If they first receive data, they work as expected. If they are reset first instead, they would get leaked: the remote uni stream is opened, but the writing-end is left open; nothing ever shuts it down; closure of the stream requires both ends to be closed; stream never closes. My changes fix this.

I have copied the same initialization code from `conn_recv_stream()` to `conn_recv_reset_stream()`.

All ngtcp2 tests pass with my change + all of my own integration tests where I use ngtcp2.

### DIFF
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -5711,11 +5711,6 @@ static int conn_recv_max_stream_data(ngtcp2_conn *conn,
       return rv;
     }
 
-    if (!bidi) {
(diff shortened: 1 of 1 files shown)


tatsuhiro-t:
Thank you for PR.  I fixed this bug in another way, see #2193

PR #2210, opened 2026-06-08
lib: Add recv_stop_sending callback

pimterry:
This adds a new optional callback, which fires when a remote STOP_SENDING frame is received.

Until now, the only way to observe this was detecting the SHUT_WR error on the next write. That's a bit limiting in itself, but also may never happen for streams that are blocked by flow control, if the application deschedules them and waits for extend_max_stream_data to resume writing. With this callback it's easy for applications to clean up stopped streams immediately instead.

I've tried to mirror the equivalent callback for received RESET_STREAM here. This is my first contribution to ngtcp2 directly, and this is a bit outside my normal area, so do let me know if I've missed something in the process here.

### DIFF
--- a/tests/ngtcp2_conn_test.c
+++ b/tests/ngtcp2_conn_test.c
@@ -324,6 +324,11 @@ typedef struct {
     uint8_t token[256];
     size_t tokenlen;
   } new_token;
+  struct {
+    int64_t stream_id;
+    uint64_t app_error_code;
+    size_t ncalled;
+  } stop_sending;
 } my_user_data;
 
 static int client_initial(ngtcp2_conn *conn, void *user_data) {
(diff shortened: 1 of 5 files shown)


tests/ngtcp2_conn_test.c
```diff
@@ -2844,11 +2865,16 @@ void test_ngtcp2_conn_recv_stop_sending(void) {
...
   ngtcp2_tpe tpe;
+  my_user_data ud;
 
   /* Receive STOP_SENDING */
   setup_default_client(&conn);
   ngtcp2_tpe_init_conn(&tpe, conn);
 
+  ud = (my_user_data){0};
+  conn->user_data = &ud;
+  conn->callbacks.recv_stop_sending = recv_stop_sending;
```
tatsuhiro-t: The recommended way to configure the custom callbacks and user data is use conn_options like:

```c
  client_default_callbacks(&callbacks);
  callbacks.recv_stop_sending = recv_stop_sending;

  opts = (conn_options){
    .callbacks = &callbacks,
    .user_data = &ud,
  };

  setup_default_client_with_options(&conn, opts);
```

pimterry: Ah, that makes sense. Now updated.

tatsuhiro-t:

PR #2276, opened 2026-08-07
Add libngtcp2_crypto_schannel, schannelclient and schannelserver

### LINKED ISSUE #1405: [FR]: schannel TLS backend
Windows 11 adds support for QUIC to `schannel`. IDK how feasible it is to make use of it in sans-io library like this, but it would be great to have it added to ngtcp2 tls framework if possible.

talregev:
libngtcp2_crypto_schannel uses Schannel TLS 1.3 QUIC message mode and CNG for packet protection on Windows.

Add CMake and Autotools integration, public API documentation, certificate-backed crypto and handshake tests, and native Windows client and server handshake examples.

The backend supports TLS_AES_128_GCM_SHA256 and
TLS_AES_256_GCM_SHA384.  It does not expose 0-RTT.

fix #1405


talregev:
@vszakats FYI
I also tested with curl (with modified changes) and it work.

vszakats:
@talregev It's a large amount of (Windows-specific) code! Did you write 
it, or is it LLM-generated?

My concern is who is going to maintain it, and fix the inevitable
CVE/bug, and how to test it? (also applies to curl) E.g. curl uses
pytest for most H3 testing and that's not working in CI. (not
necessarily impossible to make it work, but it seems to hit a lot of
[quirks](https://github.com/curl/curl/pull/20007#issuecomment-3667609782) and falls out of the portable track we use now.)

talregev:
@vszakats 
It LLM-generated.
I cannot maintain this code.
I can leave this code for good start reference for other that can maintain the code.
How we do h3 curl tests in the ci?

talregev:
@vszakats 
Do you know someone that will want to continue my work?
The implication if it will be available in release, Microsoft may use this with their curl on windows and will enable http3.
So I think it way to ship http3 on every windows. I think it a big deal here and a real change if we want to push forward http3.
What do you think?

vszakats:
I don't know such person. curl is already strained by non-portable,
Windows-specific code, in particular SSPI and Schannel. These are
underdocumented and major contributors lack the tools, will or insight
to tackle issues, maintenance and vulnerability reports (refs:
https://mastodon.social/@bagder/117056456462413525,
https://mastodon.social/@bagder/117084548583663360). Doubling the amount
of this code, esp. with LLM-generated one, that no human understands
in the first place, makes this I think, unpractical, and if you ask me,
undesired, in curl at least. It also only adds these new features to the
latest versions of Windows (though this PR misses any specifics about
it). curl still supports versions back to Vista. Testing as I said is
also an entire different issue, not currently possible, in curl at
least. There are also no Windows users in the curl team. If someone
asked, I'd recommend to use a tried and tested HTTP3 build instead.
This is what curl-for-win does.

Thinking of Windows, we'd be happy to receive help in some long-running
issues we see with _existing code_, e.g. the one where Windows OS 
decides to not return CA roots via its documented API
(https://github.com/curl/curl/issues/12303,
https://github.com/curl/curl/issues/20897), or with Unicode
(https://github.com/curl/curl/discussions/22348,
https://github.com/curl/curl/pull/22549), or with the ever-flaky Windows
runners, with broken pipes (MSYS2/Cygwin), hanging Perl
(https://github.com/curl/curl/discussions/14854), or restoring the
performance of Windows 2022 in Windows 2025 (as seen in GitHub runners
being 20-25% slower and also lacking a fast `D:` drive, making installs
slower), or other rare auth issues such as
https://github.com/curl/curl/issues/22466, with Kerberos in an 
enterprise environment. The undisablable MSBuild error auto-catch
heuristics. The lack of caching in vcpkg that limits what we can
practically test in CI with MSVC. Maybe Microsoft could jump in to help
there?

TL;DR I don't see the free capacity around the curl projects that is
ready to absorb and take over more platform-specific code, while
supporting Windows already feels like an uphill challenge.

tatsuhiro-t:
In general, I do not merge a PR that I cannot maintain.  Whether someone can help or not does not change this policy.  This is because in OSS project people come and go.  Ultimately, the maintainer needs to understand the code and fix or update it.
I have very bad experience with the proprietary TLS stacks like schannel and apple TLS.  They are very hard to maintain, and their documentations are not really nice at all, and its capability is basically depend on the operating system version and they are not free.

talregev:
I really appreciate you taking the time to think carefully about the long-term maintenance implications, especially given your previous experience maintaining proprietary TLS backends.

I understand the concern, and I agree that Schannel will inevitably add some long-term maintenance cost.

I would like to explore whether the implementation can be designed so that maintaining it remains **manageable and sustainable over time**, rather than becoming prohibitively difficult.

For example, I can look at reducing and isolating the Schannel-specific code, documenting the interaction with the Schannel APIs and Windows version dependencies, and adding Windows CI coverage that can detect regressions as early as possible.

I think this is worth the effort because native Schannel support could remove an important barrier to broader HTTP/3 support on Windows.

Would you be open to discussing what would need to change in the current implementation for its long-term maintenance burden to become acceptable to you?

tatsuhiro-t:
> Would you be open to discussing what would need to change in the current implementation for its long-term maintenance burden to become acceptable to you?

Frankly, no.  I am not interested in windows stuff and LLM generated code that even the author cannot maintain.  I'd rather suggest you to close this PR because I won't merge this.

PR #2279, opened 2026-08-12
qlog: Write CONNECTION_CLOSE reason and trigger

hunterinvariants:
Complete the existing CONNECTION_CLOSE qlog handling by writing the reason phrase and transport trigger frame type.

Reason phrases are escaped as JSON strings, and their encoded size is checked before writing to the fixed qlog buffer. Application CONNECTION_CLOSE frames omit the transport-only trigger field. Tests cover regular and escaped reasons, application closure, and insufficient buffer space.

Verified with CMake/GCC, CMake/Clang ASan+UBSan, and Autotools/GCC with warnings treated as errors.

### DIFF
--- a/lib/ngtcp2_qlog.c
+++ b/lib/ngtcp2_qlog.c
@@ -55,6 +55,80 @@ static uint8_t *write_string_impl(uint8_t *p, const uint8_t *data,
 #define write_string(DEST, S)                                                  \
   write_string_impl((DEST), (const uint8_t *)(S), ngtcp2_strlen_lit(S))
 
+static size_t escaped_stringlen(const uint8_t *data, size_t datalen) {
+  size_t len = 0;
+  size_t i;
+  size_t n;
+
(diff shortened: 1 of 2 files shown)


hunterinvariants:
Friendly ping: this PR is ready for review. Is there anything else you would like me to test, clarify, or adjust?

hunterinvariants:
Could you take a look at the proposed qlog completion for CONNECTION_CLOSE when convenient? The patch covers escaped reason phrases, the transport trigger field, application closures, and fixed-buffer limits. Please let me know if the event shape or test coverage should be changed to match the project conventions.

hunterinvariants:
@tatsuhiro-t, this has been awaiting review since August 12. The change completes the CONNECTION_CLOSE qlog output with bounded reason encoding, transport-trigger handling, and regression coverage, and the submitted checks are passing. Could you review #2279 and identify any changes required to move it forward?

PR #2283, opened 2026-08-15
Add callback for extendmaxdata

### LINKED ISSUE #2243: Why is there no callback for static conn_recv_max_data into application?
I am currently investigating a deadlock using a bidi webtransport echo stream (echo on the server and client is sending and receiving both ngtcp2 and nghttp3, node.js).  Until now, I do not know the cause.
However, I stumbled upon, that there is no callback for the app if  conn_recv_max_data(ngtcp2_conn *conn, const ngtcp2_max_data *fr) is called.
So the maximum data of the session is extended.

Imagine the following scenario, the stream max data is not exhausted, but the other max data is (and stream etc are blocking because of this=.
The other side sends an extend max data,. It is noted in the lib, but the application has no reason to reschedule all the streams and does not attempt to send data and the connection freezes.

I am not sure if this is the case, but I think it may be likely.

(@jasnell @pimterry probably relevant for the node.js implementation)

martenrichter:
Some users of the library (node.js),
require a callback, that signals, when new
maxdata values are available, as the maxdata
value is cached and evaluated outside ngtcp2.

Fixes #2243

### DIFF
--- a/lib/includes/ngtcp2/ngtcp2.h
+++ b/lib/includes/ngtcp2/ngtcp2.h
@@ -3721,12 +3721,28 @@ typedef int (*ngtcp2_stream_close2)(ngtcp2_conn *conn, uint32_t flags,
                                     uint64_t tx_app_error_code, void *user_data,
                                     void *stream_user_data);
 
+/**
+ * @functypedef
+ *
+ * :type:`ngtcp2_extend_max_data` is a callback function which
+ * is invoked when max session data is extended.
+ * |max_data| is a cumulative number of bytes
+ * an endpoint can send on this session.
+ *
+ * The callback function must return 0 if it succeeds.  Returning
+ * :macro:`NGTCP2_ERR_CALLBACK_FAILURE` makes the library call return
+ * immediately.
+ */
+typedef int (*ngtcp2_extend_max_data)(ngtcp2_conn *conn,
+                                      uint64_t max_data, void *user_data);
+
 #define NGTCP2_CALLBACKS_V1 1
 #define NGTCP2_CALLBACKS_V2 2
 #define NGTCP2_CALLBACKS_V3 3
 #define NGTCP2_CALLBACKS_V4 4
 #define NGTCP2_CALLBACKS_V5 5
-#define NGTCP2_CALLBACKS_VERSION NGTCP2_CALLBACKS_V5
+#define NGTCP2_CALLBACKS_V6 6
+#define NGTCP2_CALLBACKS_VERSION NGTCP2_CALLBACKS_V6
 
 /**
  * @struct
@@ -4088,6 +4104,17 @@ typedef struct ngtcp2_callbacks {
    * .. version-added:: 1.25.0
    */
   ngtcp2_stream_close2 stream_close2;
+  /* The following fields have been added since
+     NGTCP2_CALLBACKS_V6. */
+  /**
+   * :member:`extend_max_data` is callback function which is
+   * invoked when the maximum offset of session data that a local
+   * endpoint can send is increased.  This callback function is
+   * optional.
+   * 
+   * .. version-added:: v1.26.0
+   */
+  ngtcp2_extend_max_data extend_max_data;
 } ngtcp2_callbacks;
 
 /**
--- a/lib/ngtcp2_conn.c
+++ b/lib/ngtcp2_conn.c
@@ -485,6 +485,23 @@ static int conn_call_extend_max_stream_data(ngtcp2_conn *conn,
   return 0;
 }
 
+static int conn_call_extend_max_data(ngtcp2_conn *conn,
+                                     uint64_t datalen) {
+  int rv;
+
+  if (!conn->callbacks.extend_max_data) {
+    return 0;
+  }
+
+  rv = conn->callbacks.extend_max_data(
+    conn, datalen, conn->user_data);
+  if (rv != 0) {
+    return NGTCP2_ERR_CALLBACK_FAILURE;
+  }
+
+  return 0;
+}
+
 static int conn_call_dcid_status(ngtcp2_conn *conn,
                                  ngtcp2_connection_id_status_type type,
                                  const ngtcp2_dcid *dcid) {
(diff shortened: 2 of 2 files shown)


martenrichter:
We discussed https://github.com/ngtcp2/ngtcp2/issues/2243 in the node.js quic meeting.
Conclusion was that we like to ask you a PR for the callback.
The reason is that the current node.js quic implementation caches the max data values in their C++/js objects and need a hint, when to invalidate the values. The alternative would be to poll the value after each transmission and this does not sounds good.
I hope that I managed to match the coding style.
Thanks!

lib/ngtcp2_conn.c
```diff
@@ -11203,6 +11232,10 @@ int ngtcp2_conn_install_rx_key(ngtcp2_conn *conn, const uint8_t *secret,
       conn->remote.pending_transport_params = NULL;
       conn_sync_stream_id_limit(conn);
       conn->tx.max_offset = conn->remote.transport_params->initial_max_data;
+      rv = conn_call_extend_max_data(conn, conn->tx.max_offset);
+      if (rv != 0) {
+        return rv;
+      }
```
tatsuhiro-t: Instead of calling this here, this should only be called at:
before https://github.com/ngtcp2/ngtcp2/blob/76a41a1bc83a639f813d1313bb1af5bc61696e70/lib/ngtcp2_conn.c#L10279
and
https://github.com/ngtcp2/ngtcp2/blob/76a41a1bc83a639f813d1313bb1af5bc61696e70/lib/ngtcp2_conn.c#L10688

and they should be only invoked when conn->tx.max_offset > 0.

lib/ngtcp2_conn.c
```diff
@@ -11250,6 +11283,10 @@ int ngtcp2_conn_install_tx_key(ngtcp2_conn *conn, const uint8_t *secret,
...
```
tatsuhiro-t: Please do not call it here.  See the above comment.

lib/ngtcp2_conn.c
```diff
@@ -11748,6 +11785,10 @@ int ngtcp2_conn_set_remote_transport_params(
     }
     conn_sync_stream_id_limit(conn);
     conn->tx.max_offset = conn->remote.transport_params->initial_max_data;
+    rv = conn_call_extend_max_data(conn, conn->tx.max_offset);
+    if (rv != 0) {
+      return rv;
+    }
```
tatsuhiro-t: Please do not call it here.  See the above comment.

lib/ngtcp2_conn.c
```diff
@@ -11904,6 +11946,11 @@ int ngtcp2_conn_set_0rtt_remote_transport_params(
 
   conn->tx.max_offset = p->initial_max_data;
 
+  rv = conn_call_extend_max_data(conn, conn->tx.max_offset);
+  if (rv != 0) {
+    return rv;
+  }
+
```
tatsuhiro-t: Please do not call it here.  See the above comment.

lib/ngtcp2_conn.c
```diff
@@ -5779,15 +5796,27 @@ static int conn_recv_max_stream_data(ngtcp2_conn *conn,
...
 
-  return conn_call_extend_max_stream_data(conn, strm, fr->stream_id,
-                                          fr->max_stream_data);
+  rv = conn_call_extend_max_stream_data(conn, strm, fr->stream_id,
+                                        fr->max_stream_data);
+  if (rv != 0) {
+    return rv;
+  }
+
+  return 0;
```
tatsuhiro-t: Revert this.  No need to change these lines.

lib/includes/ngtcp2/ngtcp2.h
```diff
@@ -4088,6 +4104,17 @@ typedef struct ngtcp2_callbacks {
...
   ngtcp2_stream_close2 stream_close2;
+  /* The following fields have been added since
+     NGTCP2_CALLBACKS_V6. */
+  /**
+   * :member:`extend_max_data` is callback function which is
+   * invoked when the maximum offset of session data that a local
+   * endpoint can send is increased.  This callback function is
+   * optional.
+   * 
+   * .. version-added::
```
tatsuhiro-t:
```suggestion
   * .. version-added:: v1.26.0
```

lib/ngtcp2_conn.c
```diff
@@ -5779,15 +5796,27 @@ static int conn_recv_max_stream_data(ngtcp2_conn *conn,
...
 }
 
 /*
  * conn_recv_max_data processes received MAX_DATA frame |fr|.
  */
-static void conn_recv_max_data(ngtcp2_conn *conn, const ngtcp2_max_data *fr) {
+static int conn_recv_max_data(ngtcp2_conn *conn, const ngtcp2_max_data *fr) {
+  int rv;
   conn->tx.max_offset = ngtcp2_max(conn->tx.max_offset, fr->max_data);
+  rv = conn_call_extend_max_data(conn, fr->max_data);
```
tatsuhiro-t: The callback should only be called when max_offset is strictly increased.

lib/ngtcp2_conn.c
martenrichter: Will remove it. I overlooked it, when I backported the changes from node.js's ngtcp2 version.

lib/ngtcp2_conn.c
martenrichter: Ok, will do.

lib/ngtcp2_conn.c
martenrichter: Ok, will do

lib/ngtcp2_conn.c
martenrichter: Will do

lib/ngtcp2_conn.c
martenrichter: Will do.

martenrichter:
Ok, I have correct the things you requested and then did a force push so that the history is clean.

lib/ngtcp2_conn.c
```diff
@@ -5786,8 +5803,18 @@ static int conn_recv_max_stream_data(ngtcp2_conn *conn,
 /*
  * conn_recv_max_data processes received MAX_DATA frame |fr|.
  */
-static void conn_recv_max_data(ngtcp2_conn *conn, const ngtcp2_max_data *fr) {
+static int conn_recv_max_data(ngtcp2_conn *conn, const ngtcp2_max_data *fr) {
+  int rv;
   conn->tx.max_offset = ngtcp2_max(conn->tx.max_offset, fr->max_data);
+
+  if (fr->max_data > conn->tx.max_offset) {
```
martenrichter, line 5809: This was stupid will correct it.

martenrichter: Corrected.

tatsuhiro-t (approved):

tatsuhiro-t:
Thank you for PR. Merged now.

PR #505, opened 2022-08-10
Adding wolfSSL support as ngtcp2 crypto library and examples.

icing:
QUIC support has been added to wolfSSL (commit fa979230050dde47f6ac084c6f3b232a5e729eb5)
and ngtcp2 bindings are done as a separate crypto library.

wolfssl has been added as option to configure.

wolfssl based example client/server are build when configured
and have been verified to work against google/cloudflare/fastly
servers.

PR #2317, opened 2026-09-15
interop: Use webtransport-h3 for WebTransport CONNECT

Yanmei-Liu:
The WebTransport interop client waits for `SETTINGS_WT_ENABLED`, but `App::submit_session_request()` still sends `:protocol=webtransport`. A server enforcing the modern native HTTP/3 binding rejects this CONNECT before application data can be exchanged.

Use `webtransport-h3`, as required by [draft-ietf-webtrans-http3-16 section 3.2](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-16#section-3.2), and remove the obsolete TODO. [Section 2.1.2](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-16#section-2.1.2) distinguishes this native binding from the capsule-based protocol that uses `webtransport`.

This replaces the hard-coded older-server workaround; it does not add legacy draft negotiation. Servers that only accept the old token may need separate version-compatible handling. The peer validation below covers XQUIC.

### Validation

- `make -j4 check`: **266/266 core tests and 18/18 example tests passed**, using Clang 19 with ASan/UBSan and AWS-LC.
- Local Docker tests used unchanged upstream `quic-interop-runner` case checks. Both rebuilt clients used the same source base, dependencies and build options; the patched build changed only the interop request object. The XQUIC server image stayed fixed.

| ngtcp2 client → XQUIC server | H | UR | US | BR | BS | DR | DS |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Published ngtcp2 image, 2026-09-14 matrix | FAIL | FAIL | FAIL | FAIL | FAIL | FAIL | FAIL |
| Unmodified source rebuild, 2026-09-15 control | FAIL | — | — | — | — | — | — |
| Same rebuild with this patch, 2026-09-15 | PASS | PASS | PASS | PASS | PASS | PASS | PASS |

`—` means not rerun in that control. Both the published-image H rerun and the unmodified-source H control received CONNECT stream error `0x10e`; the patched client received HTTP 200. H verified application protocol negotiation, and all six transfer cases verified received file contents. The nine new runs completed within the runner timeout; all 18 packet captures were readable.

Reproduction inputs:

- ngtcp2 base: `50ea324d8389c0c252553d43147b3c92784ff6d7` (`webtransport`).
- nghttp3: `ffc6cdb93eb3ddb9a8f10abbd032255c5f320e1c`; AWS-LC: `47389586f8aa77c83245173793f4d44ed1d6c3a8` (v1.71.0).
- Runner: `quic-interop/quic-interop-runner@1d6f6554ac8e3b7c4cca0e3d40fd69f661420673`.
- XQUIC: `ghcr.io/alibaba/xquic/xquic-webtransport-interop@sha256:92deb6e4223d1af2aa8624adaf96dab370f92c8699fd11c076d39bc8f2ea48cc`.
- Published ngtcp2 baseline: `ghcr.io/ngtcp2/ngtcp2-interop-webtransport@sha256:9595ddc77a773da0df661af05479a408ad482593659865364af8c1b39614f907`.

The local simulator wrapper only waits for dumpcap writers at shutdown; it was identical for both controls and the patched run. These patched-client results do not represent an updated published ngtcp2 image.


tatsuhiro-t:
This is deliberate choice because when I wrote this client, I got connection rejection in webtransport interop runner.  If I confirm that webtransport-h3 works, I will change it.
This is not a huge issue because the client is targeted for interop only.  It is not a general purpose something.
Also note that we do not accept PR for other than main.

Yanmei-Liu:
Thanks for explaining the background. I understand that this is an interop-only client, and noted that PRs must target `main`.

In our latest local WebTransport interop runner tests, the XQUIC client successfully established sessions using `webtransport-h3` with **webtransport-go, Flupke, ngtcp2, and XQUIC servers**. Separately, changing this token in the ngtcp2 client made all seven cases pass against XQUIC. The results are documented in [[runner PR #505](https://github.com/quic-interop/quic-interop-runner/pull/505)](https://github.com/quic-interop/quic-interop-runner/pull/505).

[[draft-16 §3.2](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-16#section-3.2)](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-16#section-3.2) explicitly requires that `:protocol` **MUST** be `webtransport-h3` for this binding, so the current token conflicts with that requirement. Given the small scope of the change and these results, I’d still recommend updating it.

Please apply the change in whichever way you prefer and close this PR. I’m happy for it to serve as a heads-up. Thanks for taking a look!

tatsuhiro-t:
It is good to hear that webtransport-h3 finally works now.  Still webtransport requirement checks are very bad state because of browsers.
Anyway thanks for this.  Will fix it.  Closing the PR.

